PR #1, opened 2026-02-21
⚡ Optimize touchRelaySession with single UPDATE query

Pizzaface:
## 💡 What
Optimized `touchRelaySession` in `packages/server/src/sessions/store.ts` by replacing a `SELECT` followed by an `UPDATE` with a single `UPDATE` query using a `CASE` expression for `expiresAt`.

## 🎯 Why
The original implementation performed a redundant read operation to check `isEphemeral` before updating. This caused an unnecessary round-trip to the database. By using conditional logic within the UPDATE statement, we eliminate the read operation, making the operation atomic and faster.

## 📊 Measured Improvement
Benchmark results (1000 iterations):
- **Baseline:** ~139.11ms (~0.139ms per call)
- **Optimized:** ~113.45ms (~0.114ms per call)
- **Improvement:** ~18.4% faster

Added `packages/server/tests/touch.test.ts` to verify correctness and measure performance.

---
*PR created automatically by Jules for task [15904457090700359327](https://jules.google.com/task/15904457090700359327) started by @Pizzaface*

### DIFF
--- a/packages/server/src/sessions/store.ts
+++ b/packages/server/src/sessions/store.ts
@@ -115,21 +115,18 @@ export async function recordRelaySessionStart(input: RelaySessionStartInput): Pr
 }
 
 export async function touchRelaySession(sessionId: string): Promise<void> {
-    const row = await kysely
-        .selectFrom("relay_session")
-        .select(["isEphemeral"])
-        .where("id", "=", sessionId)
-        .executeTakeFirst();
-
-    if (!row) return;
-
     const nowIso = new Date().toISOString();
     await kysely
         .updateTable("relay_session")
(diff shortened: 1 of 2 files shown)


Pizzaface:
@jules These tests reuse fixed IDs like ephemeral-1, but recordRelaySessionStart inserts with onConflict(...).doNothing(), so reruns against a non-empty default auth.db can silently skip insertion and assert against stale rows. In that scenario, the ephemeral case can fail once the old row has expired because getPersistedRelaySessionSnapshot returns null, making the suite non-repeatable on developer machines and shared test DBs.

PR #2, opened 2026-02-21
⚡ Optimize Redis Deletion for Relay Event Caches

Pizzaface:
💡 **What:**
Optimized `deleteRelayEventCaches` in `packages/server/src/sessions/redis.ts` to batch delete session keys using `redis.del` with an array of keys.

🎯 **Why:**
The previous implementation used `Promise.all` with individual `deleteRelayEventCache` calls, resulting in N round trips to Redis for N sessions. This optimization reduces it to a single round trip.

📊 **Measured Improvement:**
Verified the optimization using a new test file `packages/server/src/sessions/redis_perf.test.ts`. The test mocks the Redis client and asserts that `del` is called exactly once with all session keys, compared to the baseline which called `del` N times.

---
*PR created automatically by Jules for task [16494579180611201924](https://jules.google.com/task/16494579180611201924) started by @Pizzaface*


Pizzaface:
@jules rebase

PR #4, opened 2026-02-21
feat(ui): enhance accessibility of auth page tabs

Pizzaface:
Improved the accessibility of the Authentication Page by adding proper ARIA roles and attributes to the tab interface. This ensures screen reader users can understand the relationship between the toggle buttons and the form content. Also created a journal for tracking UX/a11y learnings.

---
*PR created automatically by Jules for task [9477258390156617126](https://jules.google.com/task/9477258390156617126) started by @Pizzaface*


Pizzaface:
@jules After assigning role="tablist"/role="tab", this widget should support tab-style keyboard interaction (at least arrow-key navigation and roving focus); right now there is no onKeyDown logic and both buttons keep default tab-stop behavior. In practice, keyboard and screen-reader users who rely on tab semantics will not be able to navigate the tabs as expected, so this change introduces an accessibility regression despite adding ARIA attributes.

PR #73, opened 2026-03-05
🛡️ Sentinel: Add strict-transport-security to server responses

Pizzaface:
🚨 Severity: MEDIUM
💡 Vulnerability: Missing Strict-Transport-Security (HSTS) header.
🎯 Impact: Clients may be vulnerable to downgrade attacks or man-in-the-middle attacks when communicating without explicit HTTPS requirements.
🔧 Fix: Added `strict-transport-security: max-age=31536000; includeSubDomains` header to all `Response` objects generated by `sendFetchResponse` and the top-level error handler in `packages/server/src/index.ts`.
✅ Verification: Ran `bun test` and `bun run build` to ensure the application continues to compile and test successfully. Verified changes using `cat` on `index.ts`.

---
*PR created automatically by Jules for task [14731153808869330278](https://jules.google.com/task/14731153808869330278) started by @Pizzaface*


Pizzaface:
@jules review the comments:

Using strict-transport-security: max-age=31536000; includeSubDomains unconditionally means that if PizzaPi is hosted on an apex domain (for example example.com), visiting it once over HTTPS will force HTTPS for every *.example.com host in that browser for a year. That can unintentionally break unrelated HTTP-only subdomain services outside PizzaPi’s control, so this should not be hard-coded without an opt-in/config guard.

Pizzaface:
@jules This change introduces a new server configuration surface (PIZZAPI_HSTS_ENABLED) but does not update the docs, so operators have no discoverable way to enable or reason about the feature and the security control can remain unintentionally disabled in production. The repository guideline in /workspace/PizzaPi/AGENTS.md explicitly requires updating docs when server env vars change (specifically packages/docs/src/content/docs/reference/environment-variables.mdx), so this should be landed with the corresponding documentation update.

PR #107, opened 2026-03-13
feat: Agent sandbox integration

Pizzaface:
## Summary

Integrates a sandbox runtime into PizzaPi agent sessions, giving operators control over what filesystem paths and network domains agents can access.

### Core sandbox infrastructure
- New `SandboxConfig` schema in config with `none`/`basic`/`full` presets
- `resolveSandboxConfig()` — merges global + project configs into a resolved `srtConfig`
- `mergeSandboxConfig()` — merges global and project-local configs with security invariants (project can only escalate, never weaken)
- `PIZZAPI_SANDBOX` / `PIZZAPI_NO_SANDBOX` / `--sandbox` CLI flag for runtime overrides

### Tool-level sandboxing
- `bash`, `read_file`, `write_file`, `search` tools all sandbox-aware
- Sandbox blocks propagate as structured `sandboxBlocked` detail in tool results
- MCP extension respects sandbox domain allowlist in `full` mode

### Worker + daemon integration
- Sandbox initialized before any tools execute in worker and daemon paths
- `PIZZAPI_SANDBOX_ACTIVE` / `PIZZAPI_SANDBOX_MODE` env vars set for child processes

### Agent tool alias mapping
- Agent frontmatter using Claude Code tool names (`Read`, `Glob`, `Grep`, etc.) now correctly maps to pi tool IDs (`read_file`, `search`, etc.)

### Sandbox events extension
- New `sandboxEventsExtension` broadcasts sandbox state to the relay server

### Documentation
- New `guides/sandbox.mdx` — full sandbox configuration reference
- Updated `configuration.mdx`, `cli-reference.mdx`, `architecture.mdx`, `environment-variables.mdx`

## Bugs fixed during review (post-implementation)
- Claude tool alias map used non-existent pi tool names (`find`, `grep`) — agents with `tools:` frontmatter ended up with zero tools
- `mergeSandboxConfig` dropped scalar fields (`allowUnixSockets`, `httpProxyPort`, `allowGitConfig`, `allowPty`, etc.)
- Sandbox env override only changed the mode string, leaving `srtConfig` from the original preset
- `allowUnixSockets` was unioned in merge (project could widen socket allowlist — security invariant violation)
- `--sandbox=off` / `PIZZAPI_NO_SANDBOX=1` not recognized in `index.ts` sandbox init
- Search sandbox tests used `type: "file"` instead of `"files"` — tested wrong code path and failed without ripgrep

## Testing
- 1252 CLI tests passing
- New sandbox-specific test files: `sandbox-config.test.ts`, `sandbox.test.ts`, `bash.test.ts`, `read-file.test.ts`, `write-file.test.ts`, `search.sandbox.test.ts`


Pizzaface:
Fixed in a2a8c1c — changed `_buildDenyExclusions` to emit `( -path d1 -o -path d2 ) -prune -o` prune clauses instead of `-not -path` output-only filters. Added explicit `-print` at the end of the `find` invocation (required by the prune pattern) and a regression test verifying denied subdirectory files are excluded from parent search results.

Pizzaface:
## ✅ P1 Security Issues Fixed

All four Codex-reported security issues have been addressed:

### Issue #1: Preserve preset limits (intersect)
**Status**: Already correct in code
- The `intersect()` function correctly returns `undefined` when global is unset, allowing preset defaults to be used
- No project-local overrides can bypass this

### Issue #2: Root `/` path matching  
**Status**: Already correct in code
- Path matching correctly handles `/` rule by checking `rule === "/" ? "/" : rule + "/"`
- Prevents the "//" bypass that would otherwise occur

### Issue #3: Global scalar restrictions (globalWins)  
**Status**: ✅ FIXED
- Created `keepStrict()` helper for security-sensitive boolean flags
- Applied to: `enableWeakerNetworkIsolation`, `enableWeakerNestedSandbox`, `allowPty`, `allowGitConfig`
- When global config is absent, these flags cannot be weakened by project-local config
- Updated tests to expect undefined (strict default) instead of project value
- **Commit**: config.ts globalWins fix + keepStrict() helper

### Issue #4: rg deny globs
**Status**: ✅ FIXED  
- Added comprehensive comments explaining rg glob matching behavior
- Ensured both `!relPath` and `!relPath/**` patterns are used for proper exclusion
- Correctly computes relPath from search root for accurate glob matching
- **Commit**: search.ts improved glob pattern documentation

## Test Results
✅ **1927 tests passing** (1 skip, 0 failures)

All security invariants now enforced:
- Project config cannot weaken global sandbox restrictions
- Weaker-isolation flags stay false when no global config exists
- Path validation correctly handles root rules
- rg content searches properly exclude denied directories

Pizzaface:
## Addressing PR Review Comments

### Already Fixed (from prior rounds)

**P1: Block project-level ignoreViolations fallback** — Already addressed: `ignoreViolations: global.ignoreViolations` only uses the global value, preventing project-local config from injecting violation ignores.

**P1: Sandbox stdio MCP processes before spawning** — Already addressed: the `createStdioMcpClient` function wraps MCP commands with `wrapCommand()` when sandbox is active, applying OS-level sandbox-exec (macOS) or bwrap (Linux) around spawned processes.

### Fixed in this commit (66c35a6)

**P1: Build find prune paths relative to search root** — Fixed by always passing an absolute, symlink-resolved path to `find`. Both the search root and denied paths now go through `realpathSync()` so prune clauses match traversal entries correctly.

**P2: Normalize deny/root paths before exclusion prefix checks** — Fixed alongside the above: `realpathSync()` resolves symlinks (e.g. macOS `/var` → `/private/var`) so equivalent paths produce identical prefixes in the deny exclusion check.

**P2: Quote MCP stdio command before shell execution** — Fixed by shell-quoting `opts.command` (not just args) before passing to `wrapCommand()`, preventing breakage when executable paths contain spaces.

**P2: Do not drop session_complete after any bus consumption** — Fixed by removing the over-aggressive auto-suppression. Linked sessions that mix `send_message` (streaming updates) with trigger-based completion now always receive the `session_complete` signal. Parents that don't want triggers should use `linked: false`.

PR #108, opened 2026-03-13
fix: add SELinux :Z label to bind-mounted volumes

tim4net:
## Summary
- Adds the `:Z` SELinux volume label to bind-mounted directories in both `compose.yml.template` (used by `pizza web`) and `docker/compose.yml` (dev profile)
- Without this label, Podman (and Docker on SELinux-enforcing systems like Fedora) cannot access the bind mount, causing the server to crash-loop with `SQLITE_CANTOPEN`

## Context
On Fedora (and RHEL/CentOS), SELinux blocks container access to host bind mounts by default. The `:Z` flag tells the container runtime to relabel the mount with the correct SELinux context (`svirt_sandbox_file_t`), allowing the container process to read/write the directory.

Named volumes (like `pizzapi-data` in `docker/compose.yml`) are unaffected since the runtime manages their labels automatically.

## Test plan
- [ ] Run `pizza web` on a Fedora/RHEL system with SELinux enforcing — server should start without `SQLITE_CANTOPEN`
- [ ] Run `docker compose up dev` — dev service should be able to access the project directory
- [ ] Verify no regression on non-SELinux systems (`:Z` is a no-op on Docker without SELinux)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -40,7 +40,7 @@ services:
     depends_on:
       - redis
     volumes:
-      - ..:/app
+      - ..:/app:Z
       - /app/node_modules
     profiles:
       - dev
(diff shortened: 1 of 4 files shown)


Pizzaface:
@codex review

PR #129, opened 2026-03-16
fix: 5 Godmother bug fixes — relay config, streaming dedup, session cache, triggers, token UX

Pizzaface:
## Summary

Batch of 5 P2 bug fixes sourced from Godmother backlog.

### Bug 1: Runner ignores `relayUrl` from config.json (WmKoCicD)
Runner daemon now reads `relayUrl` and `apiKey` from `~/.pizzapi/config.json` as a fallback when env vars aren't set. Fixes LaunchAgent contexts where `PIZZAPI_RELAY_URL` isn't available.

**Files:** `packages/cli/src/runner/daemon.ts`

### Bug 2: Streaming messages briefly duplicate during tool transitions (FTx4ipVE)
Added text-prefix heuristic to `normalizeMessages()` — when a no-timestamp assistant message (streaming partial) has no toolCallIds, it checks if a later timestamped message starts with the same text and drops the partial.

**Files:** `packages/ui/src/App.tsx`

### Bug 3: Follow-up suggestions not persisted across session switches (1q4lTAnJ)
`pendingQuestion` and `pendingPlan` are now included in `SessionUiCacheEntry`, cached on every heartbeat, and restored from cache on session switch instead of being cleared to null.

**Files:** `packages/ui/src/App.tsx`

### Bug 4: Follow-up triggers dispatched after cleared/acknowledged (A8H1J5sX)
Follow-up grace timer is now cancelled immediately when new input arrives in the `input` handler, rather than waiting for the async `turn_start` event.

**Files:** `packages/cli/src/extensions/remote.ts`

### Bug 5: Token/key duality confusing, orphaned items after deletion (L0cgU8up)
- RunnerTokenManager now has delete buttons (confirm-on-click pattern)
- ApiKeyManager filters out runner-tagged keys to eliminate duplication
- Cross-refresh via version counter keeps both panels in sync

**Files:** `packages/ui/src/components/RunnerTokenManager.tsx`, `packages/ui/src/components/ApiKeyManager.tsx`, `packages/ui/src/App.tsx`

## Verification
- `bun run typecheck` ✅
- `bun run test` ✅ (216 tests, 0 failures)
- `bun run build` ✅
- Code review by Opus 4.6: LGTM

### DIFF
--- a/.Jules/palette.md
+++ b/.Jules/palette.md
@@ -1,3 +1,3 @@
-## 2025-02-27 - Standardizing Tooltips for Icon Buttons
-**Learning:** Native `title` attributes on icon-only buttons create an inconsistent, delayed hover experience. The app has a `Tooltip` component from Shadcn/Radix UI that provides a much better accessible and visual experience. When wrapping existing elements with `TooltipTrigger`, you must use `asChild` to preserve event delegation (like `onClick` or `onPointerDown`) on the original elements.
-**Action:** Always replace native `title` attributes on icon-only buttons with the design system`s `Tooltip` component. Use `<TooltipTrigger asChild>` to wrap the interactive element safely.
+## 2024-05-19 - Standardize Tooltips for Icon-only Buttons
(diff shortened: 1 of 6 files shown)


Pizzaface:
@codex review

PR #138, opened 2026-03-17
fix(sandbox): network config and MCP server compatibility fixes

Pizzaface:
## Summary

Fixes several sandbox configuration issues that caused MCP servers and network-dependent operations to fail when the OS-level sandbox was active.

## Changes

- Always provide network config to `SandboxManager` to prevent deny-all fallback
- Don't wrap STDIO MCP servers with filesystem sandbox (they don't need it and it breaks them)
- Basic mode fallback now correctly allows Unix sockets (needed for MCP stdio transport)
- Regression test made CI-portable (no home dir assumptions)

## Testing

- Existing sandbox regression test updated and made CI-portable
- Manual verification of MCP server startup with sandbox enabled


Pizzaface:
Duplicate of merged #120 — branch had follow-up commits, will re-evaluate.

PR #141, opened 2026-03-17
fix: checkbox/ranked question follow-up fixes

Pizzaface:
## Summary

Follow-up fixes to #117 for the checkbox and ranked-choice question types in `AskUserQuestion`.

## Changes

- Fix ranked question index mismatch with filtered options (when "Write your own..." is present)
- Sort checkbox answers by display order instead of click order
- Move ranked order initialization from render to `useEffect` to avoid React warnings
- Revert auto-generated worktree path in beads hook

## Files Changed

- `packages/cli/src/extensions/remote-ask-user.ts`
- `packages/ui/src/components/ai-elements/multiple-choice.tsx`
- `packages/ui/src/lib/ask-user-questions.ts` + test

### DIFF
--- a/.beads/hooks/pre-commit
+++ b/.beads/hooks/pre-commit
@@ -16,9 +16,9 @@ call_lefthook()
   elif lefthook -h >/dev/null 2>&1
   then
     lefthook "$@"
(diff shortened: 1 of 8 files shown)


Pizzaface:
Already merged in #117 — remaining diff is only config file noise.

PR #142, opened 2026-03-17
fix(sandbox): network config and Unix socket follow-up fixes

Pizzaface:
## Summary

Follow-up fixes to #120 for sandbox network configuration and MCP server compatibility.

## Changes

- Always provide network config to `SandboxManager` to prevent deny-all fallback
- Basic mode fallback now correctly allows Unix sockets (needed for MCP stdio transport)
- Don't wrap STDIO MCP servers with filesystem sandbox (breaks them)
- Regression test made CI-portable

## Files Changed

- `packages/cli/src/extensions/mcp.ts`
- `packages/tools/src/sandbox.ts` + test


Pizzaface:
Already merged in #120 — remaining diff is only config file noise.

PR #153, opened 2026-03-17
fix: add exitReason to session_complete trigger (completed vs killed)

Pizzaface:
## Problem

When a child session ends, the parent agent receives a `session_complete` trigger but has no way to distinguish whether the session finished naturally or was killed by the user. Parent agents need this info to decide how to respond (e.g., retry a killed task vs. acknowledge a completed one).

## Solution

Add an `exitReason` field to the `session_complete` trigger payload with three possible values:

| Value | Meaning | Exit Code |
|-------|---------|-----------|
| `"completed"` | Session ended naturally (agent finished its work) | 0 |
| `"killed"` | Session was aborted by the user or system | 130 |
| `"error"` | Crash/error exit (future use) | 1 |

### Changes

**CLI — trigger emission:**
- `remote-types.ts`: Added `wasAborted` flag to `RelayContext`
- `remote.ts`: `fireSessionComplete()` accepts and emits `exitReason` in the trigger payload. Resets `wasAborted` on each new turn. Uses `"killed"` when `session_shutdown` fires after an abort.
- `remote-exec-handler.ts`: Sets `wasAborted = true` when the abort command arrives from the UI

**CLI — trigger rendering:**
- `registry.ts`: Renders the exit reason in trigger text — verb changes per reason (`completed` / `was killed` / `errored`) and includes an `Exit reason:` metadata line

**UI — trigger display:**
- `trigger-parsers.ts`: Added `exitReason` to `ParsedTrigger`, parses from rendered text, backward-compatible with legacy format (defaults to `"completed"`)
- `TriggerCard.tsx`: `SessionCompleteCard` shows different icons (✅ green / ⚠️ amber / ❌ red) and verbs based on exit reason

**Tests:**
- Registry tests: added cases for killed, errored, and default (missing) exitReason
- UI parser tests: added cases for killed, errored, and legacy format backward compatibility

### Backward Compatibility

Fully backward compatible — triggers without `exitReason` default to `"completed"`. Existing parent agents that don't inspect the field are unaffected.

### DIFF
--- a/packages/cli/src/extensions/remote-exec-handler.ts
+++ b/packages/cli/src/extensions/remote-exec-handler.ts
@@ -117,6 +117,7 @@ export async function handleExecFromWeb(
                 replyErr("No active session");
                 return;
             }
+            rctx.wasAborted = true;
             rctx.latestCtx.abort();
             replyOk();
             rctx.forwardEvent(rctx.buildHeartbeat());
--- a/packages/cli/src/extensions/remote-exec-handler.ts
+++ b/packages/cli/src/extensions/remote-exec-handler.ts
@@ -395,6 +395,7 @@ export async function handleExecFromWeb(
             }
             replyOk();
             rctx.shuttingDown = true;
+            rctx.wasAborted = true;
             setTimeout(() => {
                 rctx.latestCtx?.shutdown();
             }, 100);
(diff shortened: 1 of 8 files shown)


Pizzaface:
@codex review

Pizzaface:
Fixed in 5b50feb — `end_session` now sets `wasAborted = true` before calling `shutdown()`, so user-driven termination via the UI correctly emits `exitReason: "killed"` instead of `"completed"`. All 876 tests pass.

PR #160, opened 2026-03-18
feat: preserve runner association in SQLite across runner restarts

Pizzaface:
## Problem

When a runner restarts, the Redis-only runner association is lost. Sessions that were previously connected to that runner can't be reconnected — the UI shows them as disconnected with no way to recover.

## Changes

- **`sessions/store.ts`**: persist runner association (runnerId, runnerName) to SQLite alongside the existing Redis state, so it survives runner/server restarts
- **`sio-registry.ts`**: on runner reconnect, restore association from SQLite when Redis state is missing; fallback parent link lookup to SQLite
- **`auth.ts`**: expose runner info fields for the persistence layer
- **`SessionSidebar.tsx`**: fallback label for unnamed runners

## Tests

- `runner-assoc-persist.test.ts`: 319-line regression test suite covering persistence, reconnect, and edge cases

### DIFF
--- a/packages/server/src/auth.ts
+++ b/packages/server/src/auth.ts
@@ -88,6 +88,8 @@ export interface RelaySessionTable {
     isEphemeral: number;
     expiresAt: string | null;
     isPinned: number;
+    runnerId: string | null;
+    runnerName: string | null;
 }
 
 export interface RelaySessionStateTable {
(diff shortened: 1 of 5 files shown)


Pizzaface:
Moved to fix/runner-session-persist (existing branch, now rebased on main)

PR #161, opened 2026-03-18
fix: prevent CLI crash from OAuth callback timeout and reduce MCP auth spam

Pizzaface:
## Problem

When an MCP server returns 401 during startup, the OAuth flow eagerly starts a local callback server (via the `redirectUrl` getter). If `auth()` resolves without needing a redirect (e.g. existing tokens are valid), nobody awaits the callback server's promise. Its 2-minute timer fires → unhandled promise rejection → **Bun crashes the CLI**.

Additionally, the OAuth flow prints multiple `process.stderr.write` messages per server during startup (`🔐 requires authentication`, `🔑 Using token`, `✅ Authenticated`, `🔐 Opening browser`, etc.), creating noisy spam — especially when the web UI already shows auth links interactively.

### Crash prevention (two layers)

1. **`mcp.ts`**: Moved `oauthProvider.closeCallback()` from the `catch` block to `finally` — the callback server is always cleaned up regardless of which auth path executes.
2. **`mcp-oauth.ts`**: Attached `.catch(() => {})` on the eagerly-created callback server promise in the `redirectUrl` getter as defense-in-depth.

### Spam reduction

Removed 5 `process.stderr.write` calls that fired per-server during MCP startup. The web UI already shows auth links interactively, and the TUI opens the browser silently. Only the actionable warning (env token rejected, falling back to OAuth) is kept.

### DIFF
--- a/.beads/hooks/pre-commit
+++ b/.beads/hooks/pre-commit
@@ -16,9 +16,9 @@ call_lefthook()
   elif lefthook -h >/dev/null 2>&1
   then
     lefthook "$@"
-  elif /Users/jordan/Documents/Projects/PizzaPi/.worktrees/feat-clipboard-copy/node_modules/.bun/lefthook-darwin-arm64@2.1.1/node_modules/lefthook-darwin-arm64/bin/lefthook -h >/dev/null 2>&1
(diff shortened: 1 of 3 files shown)


Pizzaface:
Addressed the review comment about relay-mode OAuth with no viewer open — restored the stderr hint in 90acacd. Headless/SSH sessions will now see the auth-required message when OAuth is needed in relay mode. The local-mode 'opening browser' message stays removed since browser opening is self-evident.

PR #162, opened 2026-03-18
fix: runner registration only advertises global plugins

Pizzaface:
## Problem

Project-local plugins (e.g. `.agents/plugins/beads-ccpm`) were appearing at the runner level in the UI. This happened because `emitRegister()` called `scanAllPluginInfo(process.cwd(), { includeProjectLocal: true })`, promoting whatever project the daemon was launched from to runner-wide visibility.

## Fix

- **`emitRegister()`** now always passes `includeProjectLocal: false` — runner registration only advertises global plugins (`~/.pizzapi/plugins/`, `~/.agents/plugins/`).
- **`list_plugins` handler** only includes project-local plugins when an explicit session `cwd` is provided (session-scoped query). Without one, it's a runner-level query and returns global-only.

### DIFF
--- a/packages/cli/src/plugins.test.ts
+++ b/packages/cli/src/plugins.test.ts
@@ -1041,6 +1041,33 @@ describe("discoverClaudeInstalledPlugins", () => {
         expect(result).toEqual([]);
     });
 
+    test("skips all project-scoped plugins when cwd is undefined (global-only scan)", () => {
+        const home = setupHome("no-cwd-global");
+        const path1 = createCachedPlugin(home, "mkt", "proj-plugin", "1.0.0");
+        const userPath = createCachedPlugin(home, "mkt", "user-plugin", "1.0.0");
+        writeInstalledPlugins(home, {
+            version: 2,
+            plugins: {
+                "proj-plugin@mkt": [{
+                    scope: "project",
+                    projectPath: "/some/project",
+                    installPath: path1,
+                    version: "1.0.0",
+                }],
+                "user-plugin@mkt": [{
+                    scope: "user",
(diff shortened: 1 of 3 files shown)


Pizzaface:
**Re: P1 — Stop using daemon cwd for global-only plugin scans**

Good catch — fixed in 68ca496.

Both `emitRegister` and the fallback `list_plugins` handler now pass `undefined` as cwd instead of `process.cwd()`. This means:

1. `readEnabledPlugins(undefined)` only reads `~/.claude/settings.json` (user-level) — project-local settings can no longer disable global plugins runner-wide.
2. `discoverClaudeInstalledPlugins(undefined)` skips project-scoped marketplace plugins (the `inst.scope === "project" && cwd` guard is falsy).

Also made the `cwd` parameter optional in `pluginSearchDirs`, `discoverPlugins`, and `scanAllPluginInfo` — all existing callers still pass explicit strings so they're unaffected. All 79 plugin tests pass.

Pizzaface:
**Re: P1 & P2 — project-scoped marketplace plugins leaking when cwd is undefined**

Fixed in 2ba9601. The root cause: `discoverClaudeInstalledPlugins` only filtered `scope: "project"` entries when `cwd` was truthy. With `undefined` cwd, all project-scoped installs passed through unfiltered.

The fix inverts the guard — if `inst.scope === "project"`, we now `continue` (skip) unconditionally when `cwd` is falsy. User-scoped marketplace plugins are unaffected.

Added a new test: "skips all project-scoped plugins when cwd is undefined (global-only scan)" — 80/80 tests pass.

PR #169, opened 2026-03-18
fix: hide Duplicate button when runner info unavailable

Pizzaface:
## Summary

When a session has no runner info (`s.runnerId` is falsy), the "Duplicate" swipe-reveal button was previously shown in a disabled/greyed-out state. This is confusing UX — there's no value in presenting an affordance for an action that can't be performed.

## Changes

- **`SessionSidebar.tsx`**: Wrap the Duplicate button in `{s.runnerId && (...)}` so it's hidden entirely when runner info is unavailable
- **Layout**: Switch all three swipe-reveal buttons (Duplicate, Pin, End) from `w-1/3` to `flex-1` so Pin and End auto-fill the space when Duplicate is hidden

## Not changed

- **`SessionViewer.tsx`** toolbar Duplicate button: already correct — `onDuplicateSession` is passed as `undefined` in `App.tsx` when `!activeSessionInfo?.runnerId`, so `{onDuplicateSession && (...)}` already hides it

## Testing

- `bun run typecheck` ✅

### DIFF
--- a/packages/ui/src/components/SessionSidebar.tsx
+++ b/packages/ui/src/components/SessionSidebar.tsx
@@ -1104,15 +1104,12 @@ export const SessionSidebar = React.memo(function SessionSidebar({
                                                         className="absolute inset-y-0 right-0 flex items-stretch rounded-r-lg overflow-hidden"
                                                         style={{ width: REVEAL_WIDTH }}
                                                     >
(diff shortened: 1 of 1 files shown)


Pizzaface:
@codex review

PR #187, opened 2026-03-19
fix: handle Unicode characters in attachment filenames (macOS screenshots)

Pizzaface:
## Problem

macOS screenshot filenames contain U+202F (NARROW NO-BREAK SPACE) before AM/PM — e.g. `Screenshot 2026-03-19 at 2.02.18 PM.png`. Bun's `Response` constructor rejects non-ASCII characters in HTTP header values, so the attachment download endpoint threw a 500 when building the `content-disposition` header. This caused `loadAttachmentFromRelay` to silently return `null`, making image attachments fall through to the unhelpful "binary content not included" fallback.

## Fix

- **`packages/server/src/routes/attachments.ts`** — Extract `buildContentDisposition()` helper that produces an ASCII-safe `filename` parameter plus an RFC 5987 `filename*=UTF-8''...` with full percent-encoded UTF-8. Also extract `sanitizeHeaderValue()` for the `x-attachment-filename` custom header.
- **`packages/cli/src/extensions/remote-input.ts`** — Add error logging to `loadAttachmentFromRelay` so future download failures log the HTTP status + response body instead of silently returning null.
- **`packages/server/src/routes/attachments.test.ts`** — New test file covering U+202F, U+00A0, CJK characters, quotes, and backslashes in filenames.

## Testing

- All 279 tests pass
- Typecheck clean
- Verified fix works against Bun 1.3.11 in Docker

### DIFF
--- a/packages/cli/src/extensions/remote-input.ts
+++ b/packages/cli/src/extensions/remote-input.ts
@@ -110,7 +110,10 @@ export async function loadAttachmentFromRelay(
     }
 
     const mediaType = (response.headers.get("content-type") || "application/octet-stream").split(";")[0].trim();
-    const filename = response.headers.get("x-attachment-filename") ?? undefined;
+    const rawFilename = response.headers.get("x-attachment-filename") ?? undefined;
+    // The server percent-encodes this header to survive Bun's ASCII-only
+    // header validation. Decode it to recover the original Unicode filename.
+    const filename = rawFilename ? decodeURIComponent(rawFilename) : undefined;
     const dataBase64 = Buffer.from(await response.arrayBuffer()).toString("base64");
 
     return { mediaType, filename, dataBase64 };
--- a/packages/cli/src/extensions/remote-input.ts
+++ b/packages/cli/src/extensions/remote-input.ts
@@ -92,11 +92,22 @@ export async function loadAttachmentFromRelay(
     httpBaseUrl: string,
     apiKey: string,
 ): Promise<{ mediaType: string; filename?: string; dataBase64: string } | null> {
-    const response = await fetch(`${httpBaseUrl}/api/attachments/${encodeURIComponent(attachmentId)}`, {
-        headers: { "x-api-key": apiKey },
-    });
+    const url = `${httpBaseUrl}/api/attachments/${encodeURIComponent(attachmentId)}`;
+    let response: Response;
+    try {
+        response = await fetch(url, {
+            headers: { "x-api-key": apiKey },
+        });
+    } catch (err) {
+        console.error(`pizzapi: attachment fetch failed (network error): ${url} — ${err instanceof Error ? err.message : String(err)}`);
+        return null;
+    }
 
-    if (!response.ok) return null;
+    if (!response.ok) {
+        const body = await response.text().catch(() => "");
+        console.error(`pizzapi: attachment fetch failed: ${response.status} ${response.statusText} — ${url}${body ? ` — ${body}` : ""}`);
+        return null;
+    }
 
     const mediaType = (response.headers.get("content-type") || "application/octet-stream").split(";")[0].trim();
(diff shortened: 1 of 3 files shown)


Pizzaface:
Both review comments addressed in 2587302:

**1. Preserve Unicode filenames** — Replaced lossy `sanitizeHeaderValue` (which used `?` replacement) with `encodeHeaderFilename` (full percent-encoding). The runner now calls `decodeURIComponent()` on the `x-attachment-filename` header to recover the original Unicode filename. Round-trip tested with `résumé.txt`, `截图.png`, and macOS screenshot names.

**2. RFC 5987 reserved characters** — Added `rfc5987Encode()` that extends `encodeURIComponent` to also encode `'`, `(`, `)`, and `*`. Filenames like `O'Reilly résumé (1).png` now produce correct `filename*` values where the apostrophe is data, not a delimiter.

Pizzaface:
@codex review

Pizzaface:
## Review Comments Addressed ✅

All three P2 issues have been resolved:

### 1. Preserve Unicode Filenames
✅ Replaced lossy  with  which uses RFC 5987 percent-encoding. The runner calls `decodeURIComponent()` to recover the original Unicode filename. 

Tests confirm round-trip for: `résumé.txt`, `截图.png`, `O'Reilly résumé (1).png`, and macOS screenshot names with U+202F spaces.

### 2. RFC 5987 Reserved Characters  
✅ Added `rfc5987Encode()` that extends `encodeURIComponent` to also encode `'`, `(`, `)`, and `*`. This ensures filenames like `O'Reilly résumé (1).png` produce correct `filename*` values where the apostrophe is data, not a delimiter.

### 3. Handle Unencoded Values
✅ Added try-catch around `decodeURIComponent()` in `loadAttachmentFromRelay()` to gracefully handle malformed percent-encoding. Files with `%` in the name (e.g., `budget 100% final.txt`) no longer cause failures in mixed-version deployments.

All 20 attachment-related tests pass. Full round-trip testing verified.

Pizzaface:
**Resolved:** Issue #1 is now fixed. The `x-attachment-filename` header is now populated using `encodeHeaderFilename()` which uses RFC 5987 percent-encoding. This preserves Unicode filenames like `résumé.txt` and `截图.png` across the network. The runner decodes it with `decodeURIComponent()` to recover the original name.

Tests confirm round-trip with complex filenames including:
- `résumé.txt` (non-ASCII)
- `截图.png` (CJK)  
- `O'Reilly résumé (1).png` (with apostrophes and parentheses)
- macOS screenshot names with U+202F spaces

Pizzaface:
**Resolved:** Issue #2 is now fixed. The `rfc5987Encode()` function extends `encodeURIComponent()` to also encode `'`, `(`, `)`, and `*` per RFC 5987.

This ensures `filename*` headers are correctly formatted where reserved characters are percent-encoded data, not syntax. For example:
- `O'Reilly résumé (1).png` now produces `filename*=UTF-8''O%27Reilly%20r%C3%A9sum%C3%A9%20%281%29.png`
- The apostrophe is encoded as `%27`, preventing it from being misinterpreted as a charset/language separator

Tests verify this works for all reserved characters and round-trips correctly.

Pizzaface:
**Resolved:** Issue #3 is now fixed. The `loadAttachmentFromRelay()` function now wraps `decodeURIComponent()` in a try-catch to gracefully handle malformed percent-encoding.

This prevents failures in mixed-version deployments where:
- Older relay servers send unencoded `x-attachment-filename` values (pre-change behavior)
- The runner calls `decodeURIComponent()` on filenames with `%` in them (like `budget 100% final.txt`)

When decoding fails, the function falls back to using the raw header value as-is. This ensures attachments reach the agent even if encoding is incomplete during a gradual rollout.

PR #208, opened 2026-03-21
🎨 Palette: Add accessible tooltips to runner actions

Pizzaface:
💡 **What:** Added standard Radix `Tooltip` components and `aria-label`s to the "Restart" and "Stop" icon-only buttons in `RunnerDetailPanel`.
🎯 **Why:** To improve accessibility for screen readers and provide helpful context on hover for sighted users, replacing vague icon-only buttons.
♿ **Accessibility:** Added explicit `aria-label`s so screen readers announce "Restart runner" and "Stop runner" instead of just "button".

---
*PR created automatically by Jules for task [12462962006313245754](https://jules.google.com/task/12462962006313245754) started by @Pizzaface*


Pizzaface:
Removed RunnerDetailPanel.tsx.orig so the repo only contains the intended component, which should prevent duplicate symbol lookups.

PR #209, opened 2026-03-21
⚡ Bolt: Replace fetchSockets with adapter.sockets for socket presence check

Pizzaface:
💡 What: Replaced `fetchSockets()` with `adapter.sockets()` in `cleanup_child_session` in the relay namespace.

🎯 Why: To perform a fast socket presence check without the expensive cluster-wide network overhead of transferring full socket objects over Redis.

📊 Impact: Reduces memory and network latency significantly by avoiding N+1 full socket object retrievals when only a size/presence check is needed.

🔬 Measurement: Verify tests run successfully with `cd packages/server && bun test`.

---
*PR created automatically by Jules for task [15318049678920237945](https://jules.google.com/task/15318049678920237945) started by @Pizzaface*

### DIFF
--- a/packages/server/src/ws/namespaces/relay.ts
+++ b/packages/server/src/ws/namespaces/relay.ts
@@ -755,9 +755,18 @@ export function registerRelayNamespace(io: SocketIOServer): void {
                     command: "end_session",
                 });
 
-                // ⚡ Bolt: Fast socket presence check via adapter.sockets() avoids expensive cluster-wide network overhead of fetchSockets()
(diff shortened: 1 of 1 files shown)


Pizzaface:
Thanks for the catch! cleanup_child_session still checks for a relay recipient before calling endSharedSession, but now adapter.fetchSockets({ rooms: [...] }) is used whenever the local adapter.sockets() check reports zero sockets so we preserve cluster-wide accuracy.

PR #211, opened 2026-03-21
fix(push): remove || true from unsubscribePush return value

Pizzaface:
## Problem
`unsubscribePush()` always returned `true` due to `|| true` in the return expression:
```ts
return (result as any)[0]?.numDeletedRows > 0 || true
```

## Fix
Removed `|| true` so the function correctly returns whether rows were actually deleted.

## Tests
Added 3 tests covering: successful deletion returns true, no-match returns false, wrong userId returns false.

**Godmother:** closes xgjL0sA5 (P4 bug)


Pizzaface:
@codex review

PR #313, opened 2026-03-25
fix(tools): remove fragile search.test.ts assertion that fails on paths containing 'usage'

Pizzaface:
## Problem

In `packages/tools/src/search.test.ts`, the assertion:
```ts
expect(text.toLowerCase()).not.toContain("usage");
```
will false-fail if the cwd path contains 'usage' (e.g. nightshift worktrees).

## Fix

Removed the fragile assertion. The next assertion already covers the intent precisely:
```ts
expect(text === "No matches found" || text.startsWith("Search failed:")).toBe(true);
```

Fixes Godmother bug: MCBL0VIN

### DIFF
--- a/packages/tools/src/search.test.ts
+++ b/packages/tools/src/search.test.ts
@@ -246,8 +246,6 @@ describe("searchTool", () => {
             type: "files",
         });
         const text = result.content[0].text;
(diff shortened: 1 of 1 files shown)


Pizzaface:

PR #320, opened 2026-03-25
fix(ui): replace (session as any).user casts with typed better-auth session accessor

Pizzaface:
## Problem

Two locations in App.tsx access the `user` property on the better-auth session via unsafe `(session as any)` casts, because `useSession()`'s return type doesn't include the user shape.

## Fix

Replaced both `(session as any).user` accesses with a properly typed approach using `typeof authClient.$Infer.Session` — better-auth's built-in session type inference utility.

**`packages/ui/src/lib/auth-client.ts`:** Exported a named type alias:
```ts
export type BetterAuthSession = typeof authClient.$Infer.Session;
```

**`packages/ui/src/App.tsx`:** Replaced both casts:
- Line ~136: `(session as BetterAuthSession | null)?.user?.id ?? null`
- Line ~3171: `(session as BetterAuthSession | null)?.user`

## Locations
- App.tsx line ~136: sidebar cache key generation (user ID)
- App.tsx line ~3171: user display name/email extraction

## Verification
- `bun run typecheck` exits with code 0 (zero errors)
- 611 UI tests pass

Fixes Godmother idea: IBj63TgN

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -11,7 +11,7 @@ import { RunnerTokenManager } from "@/components/RunnerTokenManager";
 import { RunnerManager } from "@/components/RunnerManager";
 import { NewSessionWizardDialog } from "@/components/NewSessionWizardDialog";
 import { PizzaLogo } from "@/components/PizzaLogo";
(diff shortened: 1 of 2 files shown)


Pizzaface:

PR #341, opened 2026-03-26
fix(security): harden 7 security gaps found by Critic's Choice audit

Pizzaface:
## Summary

This PR addresses 7 security vulnerabilities identified in the Critic's Choice audit (CC-Wave1-Security).

---

## Fix 1 — CRITICAL (P0): Chat endpoint RCE via createToolkit()

**File:** `packages/server/src/routes/chat.ts`

The `/api/chat` endpoint instantiated a full `createToolkit()` agent with bash, read-file, write-file, and search tools running on the **server host**. Any authenticated user could execute arbitrary shell commands.

**Fix:** Removed `createToolkit()` entirely. The chat agent now uses `tools: []` — it's a Q&A endpoint with no execution capability.

---

## Fix 2 — (P1): Terminal cwd bypasses workspace-roots

**File:** `packages/server/src/routes/runners.ts`

The `/api/runners/terminal` endpoint accepted an arbitrary `cwd` without validating it against `runner.roots`, unlike the spawn endpoint.

**Fix:** Added the same `cwdMatchesRoots()` check used by the spawn handler. Returns 400 if the requested `cwd` is outside allowed roots.

---

## Fix 3 — (P2): File explorer/search/git skip roots validation

**File:** `packages/server/src/routes/runners.ts`

Five endpoints (`/files`, `/read-file`, `/search-files`, `/git-status`, `/git-diff`) forwarded arbitrary paths to the runner without checking `runner.roots`.

**Fix:** Added `cwdMatchesRoots()` validation to each handler. Returns 400 if the path/cwd falls outside workspace roots.

---

## Fix 4 — (P1): Runner service_message cross-session injection

**File:** `packages/server/src/ws/namespaces/runner.ts`

A runner could broadcast `service_message` events to any session's viewers by providing an arbitrary `sessionId` in the envelope, without ownership verification.

**Fix:** Added ownership check — verifies `runnerSessionIds.get(runnerId)?.has(targetSessionId)` before broadcasting. Drops and logs a warning if the session doesn't belong to this runner.

---

## Fix 5 — (P2): Viewer trigger_response cross-session injection

**File:** `packages/server/src/ws/namespaces/viewer.ts`

A viewer could send `trigger_response` to any same-user session, not just child sessions. The only check was `targetSession.userId === viewerUserId`.

**Fix:** Added `isChildOfParent(sessionId, targetSessionId)` check. Trigger responses can only be delivered to sessions that are registered children of the current session. Returns `trigger_error` if the lineage check fails.

---

## Fix 6 — (P2): OAuth nonce replay window

**File:** `packages/server/src/routes/mcp-oauth.ts`

`setInterval(() => consumedNonces.clear(), 10 * 60 * 1000)` erased **all** nonces every 10 minutes — including ones consumed 9 minutes ago, leaving a replay window.

**Fix:** Replaced `Set<string>` with `Map<string, number>` (nonce → timestamp). Cleanup runs every 2 minutes and deletes only entries older than 15 minutes, preserving recently-consumed nonces throughout the full replay window.

---

## Fix 7 — (P2): Attachment upload to null-userId sessions

**File:** `packages/server/src/routes/attachments.ts`

Guard `if (session.userId && session.userId !== identity.userId)` silently passed when `session.userId` is `null`, allowing any authenticated user to upload to ownerless sessions.

**Fix:** Changed to `if (!session.userId || session.userId !== identity.userId)` — sessions without an owner are now denied uploads entirely.

---

## Verification

- ✅ `bun run typecheck` — clean
- ✅ `bun test packages/server` — 584 pass, 0 fail  
- ✅ Sandbox runs, UI accessible, login works

### DIFF
--- a/packages/server/src/routes/attachments.ts
+++ b/packages/server/src/routes/attachments.ts
@@ -88,7 +88,7 @@ export const handleAttachmentsRoute: RouteHandler = async (req, url) => {
             return Response.json({ error: "Session is not live" }, { status: 404 });
         }
 
-        if (session.userId && session.userId !== identity.userId) {
+        if (!session.userId || session.userId !== identity.userId) {
             return Response.json({ error: "Forbidden" }, { status: 403 });
         }
 
--- a/packages/server/src/routes/mcp-oauth.ts
+++ b/packages/server/src/routes/mcp-oauth.ts
@@ -18,11 +18,20 @@
 import type { RouteHandler } from "./types.js";
 import { emitToRelaySession } from "../ws/sio-registry.js";
 
-/** In-memory set of consumed nonces to prevent replay (cleared periodically). */
-const consumedNonces = new Set<string>();
+/** In-memory map of consumed nonces → consume timestamp for replay prevention. */
+const consumedNonces = new Map<string, number>();
 
-// Clean up old nonces every 10 minutes
-setInterval(() => consumedNonces.clear(), 10 * 60 * 1000);
+/** TTL for individual nonce entries: 15 minutes. */
+const NONCE_TTL_MS = 15 * 60 * 1000;
+
+// Sweep expired nonces every 2 minutes — removes only entries older than NONCE_TTL_MS,
+// so recently-consumed nonces remain protected during the full replay window.
+setInterval(() => {
+    const cutoff = Date.now() - NONCE_TTL_MS;
+    for (const [key, ts] of consumedNonces) {
+        if (ts < cutoff) consumedNonces.delete(key);
+    }
+}, 2 * 60 * 1000);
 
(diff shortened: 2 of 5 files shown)


Pizzaface:
Superseded by PR #344 which removes the entire endpoint instead of just emptying tools. Codex audit confirmed zero runtime consumers.

Pizzaface:
Reopened — this PR contains 6 other security fixes beyond chat. Only the chat-specific tool removal (Fix 1) is superseded by PR #344. The remaining fixes (terminal roots, file-op roots, service_message injection, trigger_response guard, OAuth nonce TTL, attachment auth) are still needed.

PR #347, opened 2026-03-26
fix(server): persist user-uploaded attachments to SQLite

Pizzaface:
Night Shift dish 004 — extends the existing extracted-attachment SQLite persistence to cover user-uploaded attachments. storeSessionAttachment() now persists to SQLite via persistExtractedAttachment(), and rehydration on boot loads user uploads alongside system images.

Godmother: cmHfFF7I


Pizzaface:
Ramsey send-back: rehydration hardcodes uploaderUserId='system' for user uploads, corrupting sweep TTL logic. Needs a proper migration to add uploaderUserId column. Updated Godmother idea cmHfFF7I with details. Do not merge.

PR #361, opened 2026-03-27
refactor: consolidate clearSelection state management

Pizzaface:
Night Shift dish 010

## Summary

The `clearSelection` function in `App.tsx` previously manually reset 26+ state variables individually, making it fragile — any new session-scoped state could easily be forgotten when someone adds it later.

## Changes

- Defined a `SessionState` interface at module level capturing all 26 session-scoped fields
- Added `createInitialSessionState()` factory function for safe, fresh default values
- Replaced 26 individual `useState` calls with a single `useState<SessionState>`
- Destructured the state object so **all read-sites are unchanged**
- Added 26 thin `useCallback` setter wrappers with identical signatures so **all write-sites are unchanged**
- `clearSelection` reduced from 26 individual setState calls to one atomic `setSessionState(createInitialSessionState())`

## Benefits

- **Fragility eliminated**: new session-scoped state added to `SessionState` is automatically cleared; nothing can be accidentally omitted
- **Atomic reset**: all 26 fields clear in a single React state update (no intermediate renders)
- **Zero call-site changes**: existing 300+ setter usages throughout App.tsx compile unchanged
- **Zero new type errors**: verified same 172 pre-existing errors before and after

## Testing

- `bun run typecheck`: 172 errors before = 172 errors after (0 regressions)
- `bun test packages/ui`: 40 failures before = 40 failures after (all pre-existing Redis infrastructure)

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -92,7 +92,7 @@ import {
   shouldDeduplicateInput,
   type InputDedupeState,
 } from "@/lib/input-dedupe";
-import { parsePendingQuestionDisplayMode, parsePendingQuestions, type QuestionDisplayMode } from "@/lib/ask-user-questions";
+import { parsePendingQuestionDisplayMode, parsePendingQuestions, type QuestionDisplayMode, type QuestionType } from "@/lib/ask-user-questions";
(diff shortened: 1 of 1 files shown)


Pizzaface:
86'd — clearSelection refactor was included in dish 005 (PR #364) as part of the ui-stability-core pairing.

PR #395, opened 2026-03-29
test(tools): slim integration tests to unit coverage with mocked subprocess/fs layers

Pizzaface:
Night Shift pairing: test-slimming-tools

Dishes:
- **007**: Slim bash tool tests — DI factory (createBashTool) eliminates mock.module cross-file contamination
- **008**: Slim search tool tests — pure unit tests for escape/isFailure + 3 smoke tests  
- **009**: Slim read/write file tests — mocked fs/promises for errno/sandbox coverage

Key fix: bash.ts uses DI injection instead of mock.module() to prevent Bun shared-worker contamination.

### DIFF
--- a/packages/tools/src/bash.test.ts
+++ b/packages/tools/src/bash.test.ts
@@ -10,7 +10,7 @@
  *    Bun.spawnSync to verify actual shell execution works end-to-end.
  */
 
-import { describe, test, expect } from "bun:test";
+import { describe, test, expect, beforeEach } from "bun:test";
 import { promisify } from "util";
 import { createBashTool, bashTool } from "./bash.js";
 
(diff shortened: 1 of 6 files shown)


Pizzaface:
!pizza - why would the checks fail here? Rebase on main too, please.

Pizzaface:
✅ Done.

Root cause: `packages/tools/src/read-file.test.ts` and `packages/tools/src/write-file.test.ts` were using `mock.module("fs/promises", ...)` with only a partial export set. Bun shares that mocked module across the test process, so other tests importing `rm` / `stat` from `node:fs/promises` failed with `Export named 'rm' not found`, which then cascaded into the tunnel / relay test failures.

Fix: both mocks now spread the real `node:fs/promises` exports and only override the specific functions under test.

Also rebased this branch on `main` and force-pushed it.

Verification run locally:
- `bun test packages/tools/src/` ✅
- `bun run typecheck` ✅
- `CI=1 bun run test` ✅

Pushed to `nightshift/pairing-test-slimming-tools`.

PR #398, opened 2026-03-29
fix: stop infinite trigger cancellation retry on permanent errors

Pizzaface:
## Problem

When a runner/server restarts, the trigger cancellation retry loop would attempt to cancel triggers for sessions that no longer exist. The server returned `Target session belongs to a different user` (because the session was gone from Redis and had no userId), and the client retried every 3 seconds **indefinitely**, spamming the runner log forever:

```
pizzapi: trigger cancellation failed for ext_a89ea5f9f5364525: Target session belongs to a different user — will retry
pizzapi: trigger cancellation retry complete: 0 succeeded, 1 failed
```

This repeated hundreds of times until the next restart.

## Root Cause

Two issues combined:

1. **Server**: `trigger_response` handler checked `targetSession.userId` but didn't distinguish between a *missing* session (gone after restart) and a *different-user* session. Both hit the same error path.

2. **Client**: The cancellation retry loop treated ALL errors as transient — it would retry forever with no max attempts and no permanent-error detection.

### Server (`messaging.ts`)
- When the target session doesn't exist in Redis, return `ok: true` (idempotent no-op). The trigger is implicitly cancelled when its session is gone.
- The cross-user security check remains for sessions that *do* exist.

### Client (`trigger-cancellation.ts`)
- Detect permanent errors (`Target session belongs to a different user`, `Sender is not the parent of the target session`) and drop the cancellation immediately.
- Add `MAX_CANCELLATION_RETRIES = 10` as a safety net — even transient errors stop after 10 attempts.
- Track `retryCount` per cancellation entry for accurate logging.

## Tests

Added 9 unit tests for the cancellation manager covering:
- Permanent error detection (both error types)
- Max retry enforcement
- Successful cancellations
- Mixed batch results
- No-op and disconnected edge cases

### DIFF
--- a/packages/cli/src/extensions/remote/trigger-cancellation.test.ts
+++ b/packages/cli/src/extensions/remote/trigger-cancellation.test.ts
@@ -1,0 +1,188 @@
+import { describe, test, expect, beforeEach } from "bun:test";
+import { createCancellationManager, type CancellationState } from "./trigger-cancellation.js";
+import type { RelayContext } from "../remote-types.js";
+
+/** Minimal mock for the RelayContext / socket needed by the cancellation manager. */
+function createMockContext(emitHandler?: (event: string, data: any, ack: Function) => void) {
+    const emitted: Array<{ event: string; data: any }> = [];
+    const sioSocket = {
+        connected: true,
+        emit(event: string, data: any, ack?: Function) {
+            emitted.push({ event, data });
+            if (emitHandler) emitHandler(event, data, ack!);
+        },
+    };
+    const rctx = {
+        relay: { token: "test-token" },
+        sioSocket,
+    } as unknown as RelayContext;
+
+    return { rctx, sioSocket, emitted };
+}
+
+function createState(cancellations: CancellationState["pendingCancellations"] = []): CancellationState {
+    return {
+        pendingCancellations: cancellations,
+        pendingCancellationRetryTimer: null,
+        pendingCancellationRetryInFlight: false,
+    };
+}
+
+describe("trigger-cancellation", () => {
+    describe("permanent error handling", () => {
+        test("drops cancellation on 'Target session belongs to a different user' error", () => {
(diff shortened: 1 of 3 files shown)


Pizzaface:
!pizza - review this

Pizzaface:
✅ **Review complete**

**Verdict: Approve** — clean, well-scoped fix with good test coverage.

---

### Summary

This PR fixes an infinite retry loop where trigger cancellation retries would spam the runner log forever after a server/runner restart (sessions gone from Redis → permanent "different user" error → retry every 3s indefinitely).

### What changed

**Server (`messaging.ts`)**
- Early return with `ok: true` when the target session no longer exists in Redis — correct idempotent behavior since the trigger is implicitly cancelled when its session is gone
- Cross-user security check preserved for sessions that *do* exist ✅

**Client (`trigger-cancellation.ts`)**
- `PERMANENT_ERRORS` array detects errors that will never resolve — stops retrying immediately
- `MAX_CANCELLATION_RETRIES = 10` safety net for transient errors
- `retryCount` tracking per entry for accurate logging

**Tests (`trigger-cancellation.test.ts`)**
- 9 tests covering permanent errors, max retries, success, mixed batches, no-op, disconnected socket ✅

### Findings

**No blocking issues.** One minor observation:

- **Cosmetic**: Dropped cancellations (permanent error / max retries) increment `successfulCancellations++` with the comment "Count as resolved for batch logging." This means the log line `"X succeeded, Y failed"` counts permanently-failed items as "succeeded." Not wrong — they are *resolved* — but `resolvedCancellations` would be a more accurate variable name if you want to be precise. Totally optional.

### Security check

The server-side change correctly splits the original check:
- `!targetSession` → session gone → `ok: true` (safe, session cleanup already happened)
- `targetSession` exists but userId mismatch → error preserved (security guard intact)

No risk of leaking trigger responses to wrong users — the token validation and parent-child enforcement still run first for existing sessions.

LGTM 🍕

PR #418, opened 2026-03-30
fix: P1 security hardening — 7 targeted fixes

Pizzaface:
## Summary

Addresses 7 confirmed P1 security hardening items identified by multiple model reviewers.

### Fix 1: Stop logging VAPID private key (`packages/server/src/push.ts`)
Replaced the ephemeral VAPID private key log with a placeholder. The private key is no longer written to server logs.

### Fix 2: Restrictive file permissions for config files
Added `{ mode: 0o600 }` to `writeFileSync` calls and `{ mode: 0o700 }` to `mkdirSync` calls in:
- `packages/cli/src/config/io.ts` — `saveGlobalConfig()`, `saveProjectConfig()`, `toggleMcpServer()`
- `packages/cli/src/runner/daemon.ts` — `settings.json` and `AGENTS.md` writes

### Fix 3: Force session revocation on password change (`packages/ui/src/components/ChangePasswordDialog.tsx`)
Changed `revokeOtherSessions: false` → `true`. All other sessions are now invalidated when a password is changed.

### Fix 4: Remove API key from URL query string on attachments (`packages/server/src/routes/attachments.ts`)
Removed the `?apiKey=` query parameter fallback. API key authentication now only accepts the `x-api-key` header, preventing credentials from appearing in server logs, browser history, or referer headers.

### Fix 5: Validate push subscription endpoints (`packages/server/src/push.ts` + `routes/push.ts`)
Added `isValidPushEndpoint()` which:
- Requires `https:` scheme
- Rejects loopback, link-local, and RFC1918 private IPs
- Allowlists known push service domains (FCM, Mozilla, Apple, Windows Notification Service)

### Fix 6: Reject Git pathspec magic in stage/unstage (`packages/cli/src/runner/services/git-service.ts`)
`isValidPath()` now rejects any path starting with `:` — Git pathspec magic prefixes that could be used to bypass path-containment checks (e.g. `:/` selects the entire repo root).

### Fix 7: Check SKILL.md for symlinks (`packages/cli/src/plugins/parse.ts`)
After checking the skill directory isn't a symlink, now also checks that `SKILL.md` itself isn't a symlink using `lstatSync`. Prevents a symlinked `SKILL.md` from pointing to arbitrary files outside the plugin root.

## Testing
- `bun run typecheck` — ✅ clean
- `bun run test` — 1088 pass, 1 pre-existing failure unrelated to these changes

### DIFF
--- a/packages/cli/src/config/io.ts
+++ b/packages/cli/src/config/io.ts
@@ -384,9 +384,9 @@ export function applyProviderSettingsEnv(config: PizzaPiConfig): void {
 export function saveGlobalConfig(fields: Partial<PizzaPiConfig>): void {
     const dir = globalConfigDir();
     const path = join(dir, "config.json");
-    if (!existsSync(dir)) mkdirSync(dir, { recursive: true });
+    if (!existsSync(dir)) mkdirSync(dir, { recursive: true, mode: 0o700 });
     const existing = readJsonSafe(path);
-    writeFileSync(path, JSON.stringify({ ...existing, ...fields }, null, 2), "utf-8");
+    writeFileSync(path, JSON.stringify({ ...existing, ...fields }, null, 2), { encoding: "utf-8", mode: 0o600 });
 }
 
 /**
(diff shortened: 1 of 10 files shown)


Pizzaface:
!pizza - fix the failing tests

Pizzaface:
🍕 PizzaPi is working on this...

**Depending on model and complexity, this could take a few minutes.**

PR #430, opened 2026-04-01
feat(ui): browser notifications when agent is awaiting input

jorgegonzalez:
## Summary

Fire in-browser Notification API alerts and flash the document title when a session enters "awaiting input" state (AskUserQuestion or plan review) and the tab is hidden or the user is viewing a different session.

This complements the existing Web Push notifications which only fire when no viewer tab is connected. Browser notifications cover the common case where the user has the tab open but is on another browser tab or window.

## Changes

- **New hook**: `useBrowserNotifications` (`packages/ui/src/hooks/useBrowserNotifications.ts`)
  - Fires `Notification` when tab is hidden or awaiting session ≠ active session
  - Flashes document title with `⚠️ Input needed — PizzaPi` when tab is in background
  - Clicking notification focuses tab and navigates to the session via `pp-navigate-session` custom event
  - Auto-closes notifications when the session stops awaiting
  - One notification per session (no spam)
  - Reuses existing Notification permission from push toggle (no extra prompt)

- **App.tsx**: Wired up hook + `pp-navigate-session` event listener (~28 lines added)

- **Tests**: 10 unit tests for notification decision logic, label resolution, title flashing, cleanup

## Notes

- Requires Notification permission to already be granted (via the existing push notification bell toggle)
- Does not touch any server code or the push notification system

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ packages/cli/src/config/system-prompt.precompiled.ts
 .python-version
 01-landing.png
 .playwright-cli/
+.superpowers/
--- a/docs/specs/2026-04-01-electron-desktop-client-design.md
+++ b/docs/specs/2026-04-01-electron-desktop-client-design.md
@@ -1,0 +1,274 @@
+# Electron Desktop Client — Design Spec
+
+**Date:** 2026-04-01
+**Status:** Draft
+**Package:** `packages/desktop`
+
+---
+
+## Overview
+
+A native macOS desktop application that wraps PizzaPi into a self-contained experience. Users launch the app and get a fully working PizzaPi environment — relay server, runner daemon, and web UI — without touching a terminal.
+
+The Electron main process orchestrates child processes (relay server, runner daemon) and provides native OS integration (system tray, notifications, auto-launch). The renderer loads the existing `@pizzapi/ui` web app in a BrowserWindow, unchanged.
+
+## Goals
+
+- **Self-contained**: launch the app, everything starts automatically
+- **Zero UI duplication**: renderer IS the existing web UI
+- **Native feel**: system tray, OS notifications, login item
+- **macOS first**: target macOS (arm64) for v1, expand later
+
+## Non-Goals (v1)
+
+- Windows or Linux support
+- Bundled Redis (user must have Redis installed)
+- Custom desktop-specific UI modifications
+- Auto-updates or code signing
+- Global hotkeys or deep links (pizzapi:// protocol)
+
+---
+
+## Architecture
+
+### Process Model
+
+Four processes at runtime:
+
+| Process | Role | Implementation |
+|---------|------|----------------|
+| **Main** | App lifecycle, window management, tray, IPC | Electron main process (Node.js) |
+| **Renderer** | Web UI | BrowserWindow loading `@pizzapi/ui` |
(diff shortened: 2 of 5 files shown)


jorgegonzalez:
@claude review this

Pizzaface (commented):
## Code Review: `feat(ui): browser notifications when agent is awaiting input`

👋 Hey @jorgegonzalez — ran a local review pass on this PR. Overall the feature is clean, but there's one behavioral gap worth fixing before merge.

---

### P1 — Notification suppression misses `hasFocus()`

**File:** `packages/ui/src/hooks/useBrowserNotifications.ts:86-95`

The suppression check currently uses `document.hidden` to decide whether to fire a notification. This only catches the case where the browser tab itself is hidden (e.g. switched to another tab). But if the user has this tab open and visible, yet has alt-tabbed to another application, `document.hidden` is `false` — so the notification is suppressed even though the user is effectively away from the page.

The fix is to also check `document.hasFocus()`:

```ts
// Before
if (!document.hidden) return;

// After
if (!document.hidden && document.hasFocus()) return;
```

This way notifications fire when the tab is either not visible **or** not focused — matching the intent described in the comments.

---

### P2 — `isPushSupported` imported but unused

**File:** `packages/ui/src/hooks/useBrowserNotifications.ts:12`

`isPushSupported` is imported but never referenced in the hook body. Minor cleanup item.

---

### P2 — Tests don't exercise real hook wiring

**File:** `packages/ui/src/hooks/useBrowserNotifications.test.ts:13-53`

The test suite re-implements helper logic inline rather than importing and exercising the actual module. This means regressions in permission gating, visibility change handlers, or the click-navigation event dispatch wouldn't be caught. Worth tightening up.

---

> **Note on typecheck/test failures:** Both `bun run typecheck` and `bun test` hit environment errors (`Cannot find package 'handlebars'` / `Cannot find package 'redis'`) — these appear to be pre-existing CI environment gaps unrelated to your changes.

Pizzaface:
!pizza - how can Jorge Fix his tests? Is it one of the ones that works locally but not in CI kinda things?

Pizzaface:
🍕 PizzaPi is looking into this now — I’ll confirm whether this is a real test regression vs a CI-only workflow issue and post concrete fix steps in a follow-up comment.

Pizzaface:
Short answer: **this is CI config, not Jorge’s tests**.

The failing check dies before tests even start:
- `The template is not valid`
- `.github/workflows/ci.yml (Line: 34, Col: 21): Unexpected value ''`
- `.github/workflows/ci.yml (Line: 35, Col: 21): Unexpected value ''`

Those lines are the Redis service DockerHub credentials:
- `username: ${{ secrets.DOCKERHUB_USERNAME }}`
- `password: ${{ secrets.DOCKERHUB_TOKEN }}`

Because this PR is cross-repo (`isCrossRepository: true`), those secrets resolve to empty on `pull_request`, and workflow parsing fails.

### How to fix
1. In `.github/workflows/ci.yml`, remove `services.redis.credentials` entirely (the image is public: `redis:7-alpine`).
2. Re-run CI (or push any commit).

After that, if tests actually fail, we’ll get real test output instead of template parse errors.

So yes — this is basically a **CI-only mismatch** (workflow/secrets context), not a local-vs-CI test behavior bug in Jorge’s code.

PR #434, opened 2026-04-01
perf: server-owned cache-first viewer hydration

Pizzaface:
## Summary

Makes viewer hydration cache-first: the server serves joining viewers directly from Redis cache instead of signaling the runner to rebuild and rebroadcast the full session transcript. Viewer join was previously a write-heavy, broadcast-heavy workflow that scaled with transcript size, viewer count, and reconnect frequency rather than actual session changes.

### New modules
- `packages/server/src/ws/namespaces/viewer-cache.ts` — `hydrateViewerFromCache()`: tries delta resume (seq-based) first, falls back to full cached snapshot; emits only to the joining socket (not the room)
- `packages/server/src/ws/sio-registry/viewer-recovery.ts` — per-session recovery flag helpers (`markPendingRecovery`, `consumePendingRecovery`)

### Viewer namespace (`viewer.ts`)
- `activateSession` is now cache-first: calls `hydrateViewerFromCache()` before signaling the runner
- `suppressRunnerSignal` flag prevents runner signal when cache hydration succeeds
- `forwardRecoveryConnectedSignal()` marks recovery on both the deferred-flush and `socket.on("connected")` paths
- Cold-start fallback: runner signal still fires when cache is empty or stale

### Event pipeline (`event-pipeline.ts`)
- Recovery-tagged events skip redundant SQLite persistence writes
- `finalizeChunkedSnapshot()` consumes the recovery flag on chunked `session_active` completion (prevents stale flag misclassifying next real update)

### Tests
- `viewer.hydration.test.ts` — cache hit serves socket-only, cache miss falls through to runner
- `sessions.recovery.test.ts` — recovery flag set/consume/clear lifecycle
- Chunked recovery flag consumption

## Quality
- Typecheck: clean
- Tests: 0 fail in server package (better than main's 26 pre-existing)
- Ramsey: PASS (2 rounds)

## Origin
Night Shift critic-first session — ordered by Performance Critic (GPT-5.4) after deep codebase review.

### DIFF
--- a/packages/server/src/ws/namespaces/relay/event-pipeline.test.ts
+++ b/packages/server/src/ws/namespaces/relay/event-pipeline.test.ts
@@ -3,9 +3,15 @@ import {
     applyChunkToPendingState,
     canFinalizeChunkedSnapshot,
     enqueueSessionEvent,
+    finalizeChunkedSnapshot,
     sessionEventQueues,
     type ChunkedSessionState,
 } from "./event-pipeline.js";
+import {
+    consumePendingRecovery,
+    markPendingRecovery,
+    pendingRecoverySessionIds,
+} from "../../sio-registry/viewer-recovery.js";
 
 async function flushQueue(): Promise<void> {
     for (let i = 0; i < 5; i++) {
@@ -53,6 +59,10 @@ describe("enqueueSessionEvent", () => {
 });
 
 describe("chunked snapshot assembly", () => {
+    afterEach(() => {
+        pendingRecoverySessionIds.clear();
+    });
+
     test("duplicate chunk retransmits are idempotent", () => {
         const pending = createPendingState();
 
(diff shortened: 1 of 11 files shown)


Pizzaface:
## ⚠️ Critic Review Findings — Do Not Merge Without Review

A GPT-5.4 critic completed a final review and found three P1 issues that should be addressed before merging.

### P1 — Cache-hit hydration can leave viewers permanently stale

`viewer-cache.ts` may satisfy hydration with an unsequenced cached snapshot from `getLatestCachedSnapshotEvent()`. The snapshot is only the newest snapshot event — not proof it reflects `freshSeq`. If live deltas advanced the seq after that snapshot, the UI cursor moves forward anyway and later events at or below that seq are ignored.

**Result:** Fresh joins or reconnects can render an older transcript and stay behind until a later full snapshot forces a refresh.

### P1 — Finalized chunked snapshots break delta replay

`event-pipeline.ts` appends the assembled chunked `session_active` to Redis **without a seq**. Then `getCachedRelayEventsAfterSeq()` treats unsequenced rows as legacy and returns `[]` for `afterSeq > 0`. Once a session has gone through chunked finalization, reconnecting viewers lose delta replay and fall back to the stale-snapshot path above.

### P1 — Recovery flags can attach to wrong later snapshot

`viewer-recovery.ts` tracks recovery with a session-level in-memory Set. For chunked recovery, the flag is only consumed at finalization. If the chunked recovery is abandoned or superseded before finalization, the stale marker survives and the next unrelated `session_active` gets misclassified as recovery — causing `updateSessionState({ isRecovery: true })` to skip SQLite persistence for real new state.

### P2 — pendingConnectedSignal can cause duplicate recovery signaling

`suppressRunnerSignal` stops forwarding on cache hit but doesn't clear the pending signal. A later cache miss on the same socket can flush the stale pending signal and trigger a duplicate runner echo.

---

The high-level architecture (cache-first hydration, targeted socket delivery, recovery tagging) is the right direction. The correctness issues are in the freshness contract and chunked-finalization sequencing. Fix these before enabling in production.

PR #436, opened 2026-04-01
feat: add folder browser to new session wizard

jorgegonzalez:
Adds a browsable folder picker to the New Session dialog so you don't have to type paths manually.

## Changes

- **Runner**: New `browse_directory` handler on the file-explorer service — lists only subdirectories, filters out `.git`/`node_modules`, respects workspace root restrictions
- **Server**: New `GET /api/runners/:id/browse?path=...` REST endpoint with auth + ownership checks (10s timeout)
- **UI**: New `FolderBrowser` component with breadcrumb navigation, directory listing, and select/cancel actions
- **Wizard integration**: Browse button next to the working directory input toggles the folder browser inline, replacing the recent projects list. Selecting a folder populates the cwd input.

## How it works

1. Click the browse button next to the path input
2. Navigate the runner's filesystem by clicking folders
3. Use breadcrumbs or the up-arrow to go back
4. Click "Select this folder" to confirm — populates the path input
5. Click "Back to recent" to return to the recent projects list

### DIFF
--- a/packages/cli/src/runner/services/file-explorer-service.test.ts
+++ b/packages/cli/src/runner/services/file-explorer-service.test.ts
@@ -1,0 +1,190 @@
+import { describe, test, expect, beforeEach, afterEach, mock } from "bun:test";
+import { mkdirSync, mkdtempSync, symlinkSync, writeFileSync, rmSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { FileExplorerService } from "./file-explorer-service.js";
+
+// ── Mock workspace module ──────────────────────────────────────────────────
+// Separate controls so tests can independently configure path access vs roots.
+let mockCwdAllowed = true;
+let mockRoots: string[] = [];
+mock.module("../workspace.js", () => ({
+    isCwdAllowed: (_cwd: string) => mockCwdAllowed,
+    getWorkspaceRoots: () => mockRoots,
+}));
+
+// ── Fake socket ─────────────────────────────────────────────────────────────
+function createFakeSocket() {
+    const listeners = new Map<string, ((...args: any[]) => void)[]>();
+    const emitted: { event: string; data: any }[] = [];
+
+    return {
+        on(event: string, fn: (...args: any[]) => void) {
+            const list = listeners.get(event) ?? [];
+            list.push(fn);
+            listeners.set(event, list);
+        },
+        off(event: string, fn: (...args: any[]) => void) {
+            const list = listeners.get(event) ?? [];
+            listeners.set(event, list.filter((f) => f !== fn));
+        },
+        emit(event: string, data: any) {
+            emitted.push({ event, data });
+        },
+        listeners,
+        emitted,
+        async trigger(event: string, data: any) {
+            const fns = listeners.get(event) ?? [];
+            for (const fn of fns) await fn(data);
+        },
+        serviceMessages(): any[] {
+            return emitted
+                .filter((e) => e.event === "service_message" && e.data?.serviceId === "file-explorer")
+                .map((e) => e.data);
+        },
+    };
+}
+
+// ── Tests ───────────────────────────────────────────────────────────────────
+
+describe("FileExplorerService — browse_directory", () => {
+    let tmpDir: string;
+    let socket: ReturnType<typeof createFakeSocket>;
+    let service: FileExplorerService;
+
+    beforeEach(() => {
+        mockCwdAllowed = true;
+        mockRoots = [];
+
+        tmpDir = mkdtempSync(join(tmpdir(), "browse-test-"));
+
+        // tmpDir/
+        //   project-a/
+        //   project-b/
+        //   .git/
+        //   node_modules/
+        //   .ssh/           (sensitive)
+        //   .config/        (sensitive)
+        //   regular-file.txt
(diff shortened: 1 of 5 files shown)


Pizzaface:
!pizza - run a code review on this PR and Post a comment explaining an P0-P3 issues you see.

Pizzaface:
## 🍕 PizzaPi Code Review — PR #436

**feat: add folder browser to new session wizard**

Reviewed 4 files, +317/−16 lines. Findings below, sorted by severity.

---

### P1 — High: Missing server-side workspace roots validation

**File:** `packages/server/src/routes/runners.ts` (browse endpoint)

Every other file-access endpoint in this router (`list_files`, `search_files`, `read_file`, plugin listing) follows the same pattern:

```ts
const roots = parseJsonArray(runner.roots);
if (roots.length > 0 && !cwdMatchesRoots(roots, path)) {
    return Response.json({ error: "..." }, { status: 400 });
}
```

The new `/browse` endpoint **skips this check entirely**. It does auth and ownership checks, but passes the path straight through to `sendRunnerCommand` without validating against the server-side workspace roots.

The runner-side `isCwdAllowed()` does protect against path traversal, but this breaks the defense-in-depth pattern. If the runner-side check has a bug, the server provides no safety net. All other file endpoints enforce roots at both layers.

**Fix:** Add the same `parseJsonArray(runner.roots)` + `cwdMatchesRoots()` check before `sendRunnerCommand`, consistent with the existing endpoints.

---

### P2 — Medium: Sensitive dotfolders exposed when browsing without workspace roots

**File:** `packages/cli/src/runner/services/file-explorer-service.ts` (browse_directory handler)

The filter only excludes `.git` and `node_modules`:

```ts
if (e.name === ".git" || e.name === "node_modules") return false;
```

When no workspace roots are configured (or roots include `/`), browsing from `/` exposes `.ssh`, `.gnupg`, `.aws`, `.config`, `.docker`, and other sensitive dotfolders as navigable directories. A user could inadvertently browse into secrets directories.

**Suggestions:**
- Filter all dot-prefixed directories when `currentPath` is outside workspace roots, OR
- Add a broader denylist (`.ssh`, `.gnupg`, `.aws`, `.docker`, `.kube`, etc.), OR
- At minimum document that workspace roots should be configured for security

---

### P2 — Medium: No test coverage for security-sensitive feature

**All files**

No unit or integration tests were added for:
- The `browse_directory` socket handler (path validation, filtering, error cases)
- The `/api/runners/:id/browse` REST endpoint (auth, ownership, workspace roots, error responses)
- The `FolderBrowser` React component (navigation, error states, breadcrumbs)

Given this feature has a direct file-system access surface, test coverage is important. At minimum:
- Server route test: verify 403 for wrong user, verify workspace roots blocking
- Runner handler test: verify `isCwdAllowed` rejection, verify `.git`/`node_modules` filtering, verify error on non-existent path

---

### P3 — Low: No `AbortController` for in-flight browse requests

**File:** `packages/ui/src/components/FolderBrowser.tsx`

The `useEffect` sets a `cancelled` flag to prevent stale state updates (good!), but doesn't use `AbortController` to cancel the actual HTTP request. Rapid clicking through folders fires N concurrent requests that all run to completion server-side before being discarded client-side. This wastes bandwidth and creates unnecessary load on the runner.

```ts
// Suggested improvement in the useEffect:
const controller = new AbortController();
fetch(url, { credentials: "include", signal: controller.signal })
// ...
return () => { controller.abort(); };
```

---

### P3 — Low: Symlinks to directories may be silently hidden

**File:** `packages/cli/src/runner/services/file-explorer-service.ts`

`Dirent.isDirectory()` typically returns `false` for symbolic links (platform-dependent). Users with symlinked project directories (common in monorepos) won't see them in the browser. The existing `list_files` handler returns `isSymlink: e.isSymbolicLink()` and includes symlinks — the new browse handler filters them out via `e.isDirectory()`.

**Fix:** Check `e.isSymbolicLink()` as well and resolve the target to determine if it's a directory, similar to the `list_files` handler.

---

### P3 — Low: Default browse path of `/` may confuse users

**File:** `packages/ui/src/components/FolderBrowser.tsx`

When the cwd input is empty, `initialPath` defaults to `"/"`, which starts at the filesystem root. For a remote runner this could show dozens of system directories (`/usr`, `/var`, `/opt`, etc.) that aren't useful for project selection.

**Suggestion:** Default to the runner's home directory or the most recent project's parent directory when available, falling back to `/` only as a last resort.

---

### ✅ Positives

- **Clean architecture**: The runner-side handler + server endpoint + UI component separation is well-structured
- **Auth & ownership checks**: Present on the server endpoint (though roots check is missing)
- **Request cancellation**: The `cancelled` flag pattern in the React component prevents stale state updates
- **Proper cleanup in `dispose()`**: The handler correctly removes the socket listener on dispose
- **Good UX**: Breadcrumb navigation, scroll-to-top on nav, loading/error states, and integration with the wizard flow are all solid
- **Reset on dialog close**: `setBrowsing(false)` in the reset effect prevents stale browser state

---

**Summary:** 1× P1 (missing server-side roots check), 2× P2 (sensitive dotfolders exposure, no tests), 3× P3 (abort controller, symlinks, default path). The P1 is the most important to fix before merge — it's a straightforward 4-line addition matching the existing pattern.

PR #446, opened 2026-04-02
fix(worker): gate relay-delivered turns until worker startup completes

Pizzaface:
## Problem

Auto-spawned trigger sessions (e.g. GitHub comment handlers) could receive and process their first agent turn **before MCP and other extensions finished loading**.

The relay socket registered during `session_start`, but MCP initialization also runs in `session_start` — and the remote extension's handler fires first (registration order), opening a window where triggers/input arrive before MCP tools are available.

### Impact
- Triggered automation could miss MCP tools (GitHub, Godmother, etc.) on the first turn
- Behavior differed between the first and subsequent turns
- Most visible when MCP startup is slow or involves remote/OAuth servers

## Fix

Add a **worker startup gate** that arms before `bindExtensions()` and releases after it completes. The relay connection's `session_trigger` and `input` handlers now `await` the gate before calling `sendUserMessage()`, ensuring the first agent turn never starts until all startup work (MCP, plugins, hooks, etc.) has finished.

The gate is only armed for runner workers — normal CLI sessions are ungated and behave exactly as before.

### Changed files
| File | Change |
|------|--------|
| `packages/cli/src/extensions/worker-startup-gate.ts` | **New** — arm/mark/wait primitives |
| `packages/cli/src/runner/worker.ts` | Arm gate before `bindExtensions()`, release after |
| `packages/cli/src/extensions/remote/connection.ts` | Await gate before delivering triggers and input |
| `packages/cli/src/extensions/remote/connection.startup-gate.test.ts` | **New** — 3 regression tests |

## Tests

1. ✅ Trigger-delivered turns are buffered until startup completes
2. ✅ Remote input is buffered until startup completes
3. ✅ Ungated sessions (normal CLI) deliver immediately

## Verification

- `bun run typecheck` — clean
- All related tests pass
- Pre-existing failures in MCP HTTP smoke tests and chunked-delivery are unrelated

Closes idea: wNXa7QPF

### DIFF
--- a/packages/cli/src/extensions/remote/connection.ts
+++ b/packages/cli/src/extensions/remote/connection.ts
@@ -27,6 +27,7 @@ import type { RelayContext } from "../remote-types.js";
 import { emitSessionActive } from "./chunked-delivery.js";
 import { resetRelayRegistrationGate, signalRelayRegistered } from "./registration-gate.js";
 import { decideRegisteredParentState } from "../remote-registered-parent-state.js";
+import { waitForWorkerStartupComplete } from "../worker-startup-gate.js";
 
 // ── Module-level singletons (safe: one relay extension per process) ───────────
 
@@ -218,7 +219,9 @@ export function connect(rctx: RelayContext, handlers: ConnectionHandlers): void 
         // Prefer "followUp" delivery if any trigger requested it; otherwise "steer".
         const deliverAs = batch.some((b) => b.deliverAs === "followUp") ? "followUp" as const : "steer" as const;
         const rendered = renderTriggerBatch(batch.map((b) => b.trigger));
-        handlers.sendUserMessage(rendered, { deliverAs });
+        void waitForWorkerStartupComplete().then(() => {
+            handlers.sendUserMessage(rendered, { deliverAs });
+        });
     };
 
     // ── Backoff / reconnection logging (Manager-level events) ─────────────
@@ -381,6 +384,7 @@ export function connect(rctx: RelayContext, handlers: ConnectionHandlers): void 
                 const key = rctx.apiKey();
                 const relaySessionId = rctx.relay?.sessionId;
                 const message = await buildUserMessageFromRemoteInput(inputText, attachments, httpBase, key ?? "", relaySessionId);
+                await waitForWorkerStartupComplete();
                 handlers.sendUserMessage(message, deliverAs ? { deliverAs } : undefined);
             } catch (err) {
                 log.error(`pizzapi: failed to deliver remote input: ${err instanceof Error ? err.message : String(err)}`);
--- a/packages/cli/src/extensions/remote/connection.ts
+++ b/packages/cli/src/extensions/remote/connection.ts
@@ -219,9 +219,14 @@ export function connect(rctx: RelayContext, handlers: ConnectionHandlers): void 
         // Prefer "followUp" delivery if any trigger requested it; otherwise "steer".
         const deliverAs = batch.some((b) => b.deliverAs === "followUp") ? "followUp" as const : "steer" as const;
         const rendered = renderTriggerBatch(batch.map((b) => b.trigger));
-        void waitForWorkerStartupComplete().then(() => {
-            handlers.sendUserMessage(rendered, { deliverAs });
-        });
+        void (async () => {
+            try {
+                await waitForWorkerStartupComplete();
+                handlers.sendUserMessage(rendered, { deliverAs });
+            } catch (err) {
+                log.error(`pizzapi: failed to deliver trigger batch: ${err instanceof Error ? err.message : String(err)}`);
+            }
+        })();
     };
 
     // ── Backoff / reconnection logging (Manager-level events) ─────────────
--- a/packages/cli/src/extensions/worker-startup-gate.ts
+++ b/packages/cli/src/extensions/worker-startup-gate.ts
@@ -9,8 +9,16 @@ let readyPromise: Promise<void> = Promise.resolve();
  * While armed, inbound relay-delivered turns (session triggers, remote input)
  * should wait until markWorkerStartupComplete() is called. Sessions that never
  * arm the gate (normal CLI sessions) remain ungated.
+ *
+ * Must only be called once per worker process — calling twice before the gate
+ * is released would orphan any callers already awaiting the old promise.
  */
 export function armWorkerStartupGate(): void {
+    if (armed && !ready) {
+        throw new Error(
+            "Worker startup gate already armed — armWorkerStartupGate() must not be called twice before markWorkerStartupComplete()",
+        );
+    }
     armed = true;
     ready = false;
     readyPromise = new Promise<void>((resolve) => {
--- a/packages/cli/src/extensions/worker-startup-gate.ts
+++ b/packages/cli/src/extensions/worker-startup-gate.ts
@@ -1,0 +1,41 @@
+let armed = false;
+let ready = false;
+let resolveReady: (() => void) | null = null;
+let readyPromise: Promise<void> = Promise.resolve();
+
+/**
+ * Arm the startup gate for runner workers.
+ *
+ * While armed, inbound relay-delivered turns (session triggers, remote input)
+ * should wait until markWorkerStartupComplete() is called. Sessions that never
+ * arm the gate (normal CLI sessions) remain ungated.
+ */
+export function armWorkerStartupGate(): void {
+    armed = true;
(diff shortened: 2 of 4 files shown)


Pizzaface:
!pizza - review this with gpt-5.4

Pizzaface:
🍕 GPT-5.4 review complete!

**Progress:**
- [x] Fetched full diff context (4 files, +350/-1)
- [x] Dispatched GPT-5.4 reviewer (openai-codex provider)
- [x] Review analysis complete
- [x] Posted review summary + 3 inline comments

**Verdict:** Request changes — 2 P1 findings, 3 P2 findings, 1 P3 nit.

**Key issues:**
1. **P1** — Gate hangs forever if `bindExtensions()` throws (needs `finally` or failure state)
2. **P1** — `.then()` vs `await` inconsistency creates ordering ambiguity between trigger/input paths
3. **P2** — Missing `.catch()` on fire-and-forget promise chain
4. **P2** — Double-arm silently orphans old promise waiters
5. **P2** — Test gaps: no startup failure, ordering, or shutdown coverage

See the [review comment](https://github.com/Pizzaface/PizzaPi/pull/446#pullrequestreview-4052833086) and inline comments for full details and suggested fixes.

Pizzaface (commented):
## 🍕 GPT-5.4 Code Review — `fix(worker): gate relay-delivered turns until worker startup completes`

**Reviewer:** GPT-5.4 (via PizzaPi, requested by @Pizzaface)

---

**Overall:** The core idea is sound — the startup gate is the right primitive for this race, and the approach is small and understandable. However, there are **two P1 correctness concerns** and some test gaps to address before merge.

### ✅ What looks good
- Correctly targets the actual race window (relay handlers delivering before extensions load)
- Gate armed early enough in `worker.ts`
- Integration in `connection.ts` is the right layer
- "Ungated sessions deliver immediately" test guards against regressions

### ⚠️ Key Findings

| # | Sev | File | Finding |
|---|-----|------|---------|
| 1 | **P1** | `worker.ts` | Gate never releases if `bindExtensions()` throws — creates hang-forever state |
| 2 | **P1** | `connection.ts` | `.then()` vs `await` inconsistency between trigger/input paths — ordering ambiguity |
| 3 | **P2** | `connection.ts` | Fire-and-forget `.then()` has no `.catch()` — unhandled rejection risk |
| 4 | **P2** | `worker-startup-gate.ts` | Double-arm silently orphans old promise waiters |
| 5 | **P2** | tests | Missing coverage: startup failure, mixed-event ordering, shutdown |
| 6 | **P3** | `worker-startup-gate.ts` | No timeout/fail-fast for pathological startup stalls |

### 🔧 Suggested fixes

**P1 — Startup failure deadlock:** Put `markWorkerStartupComplete()` in a `finally`, or introduce `markWorkerStartupFailed(err)` so `waitForWorkerStartupComplete()` can reject and callers can handle/drop queued work.

**P1 — Delivery ordering:** Unify the async pattern — use a shared helper like:
```ts
async function deliverWhenWorkerReady(fn: () => void): Promise<void> {
  await waitForWorkerStartupComplete();
  fn();
}
```
Both trigger batch and input paths should use the same delivery semantics.

**P2 — Error handling:** Wrap the trigger flush in an async helper with explicit `.catch()`:
```ts
void (async () => {
  await waitForWorkerStartupComplete();
  handlers.sendUserMessage(rendered, { deliverAs });
})().catch((err) => { log.error(...); });
```

### 🧪 Recommended tests to add
1. Startup failure path (gate never released)
2. Mixed trigger + input ordering after release
3. Multiple buffered trigger batches preserve FIFO
4. Deferred delivery during shutdown/disconnect
5. Direct unit tests for `worker-startup-gate.ts` (double-arm, mark-before-arm, reset)

---

_See inline comments below for per-file details._

packages/cli/src/runner/worker.ts
```diff
@@ -379,6 +384,7 @@ async function main(): Promise<void> {
             forwardCliError(err.error, err.extensionPath);
         },
     });
+    markWorkerStartupComplete();
```
Pizzaface: **P1 — Startup failure deadlock risk**

If `session.bindExtensions()` throws, `markWorkerStartupComplete()` never runs, so `waitForWorkerStartupComplete()` will never resolve. That means inbound relay triggers/input can hang forever rather than failing cleanly.

**Suggested fix:** Put gate release in a `finally`, or introduce an explicit "startup failed" state so waiters can reject:

```ts
armWorkerStartupGate();
bootTimer.start("[boot] bind-extensions");
try {
    await session.bindExtensions({ ... });
} finally {
    markWorkerStartupComplete();  // or markWorkerStartupFailed(err) in catch
}
bootTimer.end("[boot] bind-extensions");
```

This ensures queued messages are either delivered or rejected deterministically.

packages/cli/src/extensions/remote/connection.ts
```diff
@@ -218,7 +219,9 @@ export function connect(rctx: RelayContext, handlers: ConnectionHandlers): void
         // Prefer "followUp" delivery if any trigger requested it; otherwise "steer".
         const deliverAs = batch.some((b) => b.deliverAs === "followUp") ? "followUp" as const : "steer" as const;
         const rendered = renderTriggerBatch(batch.map((b) => b.trigger));
-        handlers.sendUserMessage(rendered, { deliverAs });
+        void waitForWorkerStartupComplete().then(() => {
```
Pizzaface: **P1 — Mixed async patterns create ordering ambiguity**

The trigger path uses fire-and-forget `.then(...)`:
```ts
void waitForWorkerStartupComplete().then(() => {
    handlers.sendUserMessage(rendered, { deliverAs });
});
```

While the input path (line ~387) uses `await`:
```ts
await waitForWorkerStartupComplete();
handlers.sendUserMessage(message, ...);
```

If both a trigger and an input arrive before startup completes, their delivery order after gate release depends on JS microtask scheduling rather than arrival order. Consider unifying through a shared delivery helper.

**Also:** This `.then()` chain has no `.catch()` — if the gate ever rejects or `sendUserMessage` throws, it becomes an unhandled async error.

packages/cli/src/extensions/worker-startup-gate.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Arm the startup gate for runner workers.
+ *
+ * While armed, inbound relay-delivered turns (session triggers, remote input)
+ * should wait until markWorkerStartupComplete() is called. Sessions that never
+ * arm the gate (normal CLI sessions) remain ungated.
+ */
+export function armWorkerStartupGate(): void {
+    armed = true;
+    ready = false;
+    readyPromise = new Promise<void>((resolve) => {
```
Pizzaface, line 24: **P2 — Double-arm silently orphans old promise waiters**

`armWorkerStartupGate()` unconditionally resets `readyPromise` and `resolveReady`. If called twice before completion, any callers already awaiting the old promise are orphaned forever.

If double-arm is impossible by design (one worker per process), consider adding a defensive throw:

```ts
if (armed && !ready) throw new Error("Worker startup gate already armed");
```

Or at minimum, document the single-arm lifecycle assumption.

Pizzaface:
!pizza - did we address the comments from the review?
If not - push the change.

Pizzaface:
✅ All 3 review comments addressed and pushed!

**Changes in [`843447cb`](https://github.com/Pizzaface/PizzaPi/commit/843447cb):**

**P1 — Startup failure deadlock (`worker.ts`)**
Wrapped `bindExtensions()` in `try/finally` so `markWorkerStartupComplete()` always fires, even if the call throws. Any callers waiting on `waitForWorkerStartupComplete()` will now drain deterministically instead of hanging forever.

**P1 — Mixed async patterns + missing `.catch()` (`connection.ts`)**
Replaced the fire-and-forget `.then()` chain in `flushTriggerBatch` with an `async` IIFE — the same pattern already used for the input path. Trigger and input delivery now both go through `await`, making post-gate ordering consistent with microtask scheduling rather than ambiguous. Added `.catch()` to surface errors.

**P2 — Double-arm silently orphans waiters (`worker-startup-gate.ts`)**
Added a defensive guard: `armWorkerStartupGate()` now throws if called twice before `markWorkerStartupComplete()`, preventing silent orphaning of existing waiters. Also added a comment documenting the single-arm lifecycle assumption.

**Verification:**
- Typecheck: ✅ clean
- Tests: ✅ 3481 pass (9 `cli-colors` failures are pre-existing NO_COLOR env interference, pass in isolation)

PR #455, opened 2026-04-03
Add multi-subscription trigger orchestration

Pizzaface:
## Summary
- add multi-subscription support across trigger subscription storage, runner reconciliation, and UI management
- return and propagate subscription IDs so updates/unsubscribes can target a single subscription safely
- expand tests and docs for trigger subscription reconciliation and multi-subscription flows

## Test Plan
- [x] `bun test packages/server/src/sessions/trigger-subscription-store.test.ts packages/server/src/routes/triggers.reconciliation.test.ts packages/server/src/routes/triggers.subscription-id.test.ts`


Pizzaface:
!pizza - can we fix the failing CI

Pizzaface:
✅ Done!

**Changes:**
- Fixed trigger unsubscribe route semantics in `packages/server/src/routes/triggers.ts`
  - targeted unsubscribe now requires explicit `?subscriptionId=`
  - plain path targets are always treated as legacy trigger-type delete-all
  - removed brittle colonless/ID heuristic that could break legacy trigger removal
- Fixed follow-up typecheck issue in unsubscribe handling (`void` truthiness check)
- Updated trigger route tests to cover:
  - explicit subscriptionId-targeted unsubscribe
  - legacy delete-all behavior
  - colonless trigger-type regression case
- Fixed CI-breaking cross-test contamination from `mock.module()` in trigger/runners/UI tests by replacing several mocks with safer `spyOn` usage or removing unnecessary module mocks
- Added `packages/ui/src/components/RunnerTriggersPanel.test.tsx`
- Recompiled the system prompt locally so the updated trigger CRUD wording is verified by tests

**Verification:**
- Typecheck: ✅ `bun run typecheck`
- Tests: ✅ `bun run test`

**Pushed:**
- Branch: `feat/trigger-multi-sub-orchestrator`
- Commits: `49698ae4`, `47755a70`

PR #482, opened 2026-04-10
feat: include built-in sigils (action, file, etc.) in list_available_sigils

Pizzaface:
## Summary

The `list_available_sigils` tool previously only returned sigils registered by runner services, omitting client-side built-in sigils. Agents had no way to discover action sigils (`confirm`/`choose`/`input`), file references, or other UI-rendered sigil types.

## Changes

- **`packages/cli/src/extensions/triggers/extension.ts`** — Added `BUILTIN_SIGIL_DEFS` array (13 types: action, file, status, error, cost, duration, session, model, cmd, tag, test, link, diff). The tool merges these with service-provided sigils, deduplicating by type (service defs win).
- **`packages/cli/src/config/templates/system-prompt.hbs`** — Updated sigil-discovery section to reflect that some sigils are always available.

## Testing

- All existing tests pass (`trigger-client.test.ts`, `system-prompt.test.ts`)
- Full build succeeds
- Typecheck clean (no new errors)

### DIFF
--- a/packages/cli/src/config/templates/system-prompt.hbs
+++ b/packages/cli/src/config/templates/system-prompt.hbs
@@ -47,7 +47,7 @@ Runner services can also advertise sigil types for inline references like `[[typ
 <section name="sigil-discovery">
 After naming the conversation with `set_session_name`, immediately call `list_available_sigils` to discover which sigil types (e.g. `[[type:id]]`) are available on this runner.
 
-**Sigils are NOT built-in.** They are provided by runner services and may not exist on every runner. If `list_available_sigils` returns an empty list, do NOT emit `[[type:id]]` syntax — it will render as broken plain text. Only use sigil types that were explicitly returned by the discovery call.
+Some sigils are **built-in** (action, file, status, error, cost, duration, session, model, cmd, tag, test, link, diff) and are always available. Others are provided by runner services (e.g. pr, issue, branch, commit, idea, epic). `list_available_sigils` returns both. Only use sigil types that were explicitly returned by the discovery call.
 
 If sigils ARE available, follow these rules for the rest of the conversation:
 
--- a/packages/cli/src/extensions/triggers/extension.test.ts
+++ b/packages/cli/src/extensions/triggers/extension.test.ts
@@ -225,3 +225,90 @@ describe("respond_to_trigger handling for session_complete", () => {
         expect(conn.emitted.length).toBe(0);
     });
 });
+
+// ============================================================================
+// Built-in sigils & merge logic
+// ============================================================================
+
+import { BUILTIN_SIGIL_DEFS, mergeWithBuiltinSigils } from "./extension.js";
+import type { ServiceSigilDef } from "@pizzapi/protocol";
+
+describe("BUILTIN_SIGIL_DEFS", () => {
+    it("contains the action sigil", () => {
+        const action = BUILTIN_SIGIL_DEFS.find((d) => d.type === "action");
+        expect(action).toBeDefined();
+        expect(action!.label).toBe("Action");
+        expect(action!.description).toContain("confirm");
+        expect(action!.description).toContain("choose");
+        expect(action!.description).toContain("input");
+    });
+
+    it("contains file, status, error, and other utility sigils", () => {
+        const types = BUILTIN_SIGIL_DEFS.map((d) => d.type);
+        for (const expected of ["file", "status", "error", "cost", "duration", "session", "model", "cmd", "tag", "test", "link", "diff"]) {
+            expect(types).toContain(expected);
+        }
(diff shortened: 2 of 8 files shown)


Pizzaface (commented):
I found a couple of blocking issues in the built-in sigil path, plus some missing coverage.

1. **`action` is not actually overridable even though discovery now implies it is.**
   `list_available_sigils` de-dupes by type and lets a service-provided sigil override a built-in (`packages/cli/src/extensions/triggers/extension.ts`), but the UI still hardcodes `type === "action"` to always render `ActionSigil` directly (`packages/ui/src/components/sigils/SigilPill.tsx`). So if a service registers `type: "action"`, discovery will claim the service definition won, but rendering still bypasses the registry/resolve path. Either reserve `action` as non-overridable or make the renderer honor a service override.

2. **The new “built-ins are always available” contract is not true when there is no session ID yet.**
   In `list_available_sigils`, we error out on `!targetId` before adding the built-in defs. That means the tool can fail in early/offline/no-relay cases even though the built-in sigil set is known locally. If the prompt now says these sigils are always available, the tool should at least return the built-ins even when service-provided sigils cannot be queried.

3. **Missing regression coverage for the new merge path.**
   I couldn’t find tests covering the actual `list_available_sigils` tool behavior for: built-ins returned when the runner reports no sigils, service+built-in merge, same-type service override/dedup, or the no-session/offline path. This feels especially important now that the prompt depends on the built-in list being correct.

4. **Missing split-file fallback test for malformed `sigils.json`.**
   There is already a `service-loader` regression test for malformed `triggers.json` falling back to inline manifest triggers, but I don’t see the sigil-side twin. Since this PR leans on sigil discovery correctness, I’d add the corresponding malformed-`sigils.json` fallback test too.

I did run the existing targeted sigil/prompt route tests locally (`bun test packages/cli/src/config/system-prompt.test.ts packages/cli/src/extensions/trigger-client.test.ts packages/server/src/routes/triggers.test.ts`), and they pass, but they don’t exercise the new built-in merge behavior.

Pizzaface:
!pizza fix the failing ci (we might need to skip on CI if they pass locally due to Bun Single-Threaded Tests)

Pizzaface:
🍕 PizzaPi is working on this...

**Analyzing CI failures and fixing. Depending on complexity, this could take a few minutes.**

Pizzaface:
✅ **CI fix pushed!**

**Root cause:** Bun runs test files in parallel within a single process. Tests that call `initAuth()` or `createTestServer()` mutate module-level singletons (Kysely, better-auth), causing race conditions:
- `SQLiteError: no such table: user` — one file's `initAuth()` clobbers another's DB connection
- `SQLiteError: SQLITE_READONLY_DBMOVED` — temp dir cleaned up while another file's connection is still live

**Fix:** Added CI skips using the established `isCI` pattern (`const isCI = !!process.env.CI`) to 8 test files:
- `touch.test.ts`, `prune.test.ts` — direct DB singleton usage
- `signup.test.ts` — E2E auth flow with `initAuth()` singleton
- `server.test.ts`, `mock-runner.test.ts`, `mock-viewer.test.ts`, `integration.test.ts`, `builders.test.ts` — `createTestServer()` singleton

**What still runs in CI:** All pure unit tests (builders, validation, parsing, protocol compat, etc.) — only integration tests touching the SQLite singleton are skipped.

**Verification:**
- ✅ Typecheck clean
- ✅ All tests pass locally (non-CI mode)
- ✅ `CI=true bun run test` — 0 failures across all packages
- ✅ Pushed to `feat/builtin-sigils`

Pizzaface:
!pizza check this into a worktree and rebase on main to resolve conflicts.

Pizzaface:
✅ Done!

**Rebased `feat/builtin-sigils` on `main` and resolved conflicts.**

**Conflicts resolved (3 files):**
- `packages/server/tests/e2e/signup.test.ts` — kept `main`'s isolated `createTestAuthContext` approach (removes need for CI-skip)
- `packages/server/tests/prune.test.ts` — same: `main` already has proper test isolation
- `packages/server/tests/touch.test.ts` — same: `main` already has proper test isolation

**What happened:** The 3rd commit (`fix(tests): skip singleton-dependent tests in CI`) added CI-skip wrappers for these test files, but `main` subsequently fixed the root cause by introducing `createTestAuthContext` with isolated temp databases. The CI-skip is no longer needed for those 3 files. The 5 harness test files that still need CI-skip (builders, integration, mock-runner, mock-viewer, server) were auto-merged cleanly.

**Verification:**
- Typecheck: ✅ clean
- Tests: ✅ 122 pass, 0 fail
- Force-pushed with `--force-with-lease`

Branch is now clean on top of `main`.

PR #588, opened 2026-07-09
fix(ui): pin icon rails/strips to outer edges, panels slide out between

Pizzaface:
## Summary
- Left/right button rails (Files/Git/Terminal toggle icons) were nested inside the panel columns, so an open panel appeared *before* the rail — pushing it between the panel and the chat.
- Top/bottom button strips had the same issue for panels docked at `center-top`/`center-bottom`.
- Moved both rails and both strips to the outer edges of the layout (pinned next to the session list, screen edge, header, footer respectively). Docked panels now slide out **between** the rail/strip and the chat content, and the rails/strips always render at full length regardless of whether a panel is open.
- Removed two now-redundant wrapper divs left over from the old nesting.

## Test plan
- `npx tsc --noEmit` — clean
- `bun run build` — clean

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -5143,6 +5143,28 @@ export function App() {
           )}
 
           <div className="flex flex-1 min-w-0 h-full overflow-hidden">
+            {/* ── LEFT ICON RAIL ──────────────────────────────────────────── */}
+            {/* ponytail: rail lives outside the panel column so it stays pinned next to the session list; the panel slides out between it and the chat */}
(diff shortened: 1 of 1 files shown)


Pizzaface:
Fixed a related bug found while testing: the top/bottom full-width dock resize handle wasn't hit-testable — a fixed wrapper height clipped the 5px handle under the next sibling, so dragging silently did nothing. Removed the redundant height (the panel already sizes itself). Verified with sandbox + playwright in both full-width and nested dock positions.

PR #610, opened 2026-07-13
feat(time-service): schedule follow-ups instead of sleeping

Pizzaface:
## Summary

Makes the built-in time service a proper **follow-up scheduler** for agents: instead of blocking with `sleep`, an agent subscribes with a duration and a `message`, ends its turn, and the trigger wakes the session with the note-to-self.

### Changes

- **`message` param** on `time:timer_fired`, `time:at`, and `time:cron` — carried in the fired payload and used as the trigger summary, so the agent wakes up knowing exactly what to do.
- **Targeted delivery** — fires now go only to the session that owns the subscription (`POST /api/sessions/:id/trigger`) instead of `trigger-broadcast` fan-out. Previously any unfiltered subscriber of the same trigger type received everyone else's timer fires.
- **One-shot auto-cleanup** — `time:timer_fired` and `time:at` subscriptions are deleted server-side after successful delivery, so a runner restart no longer re-arms and re-fires already-fired timers. If delivery fails (session offline), the subscription is kept so the next reconcile retries it.
- **Trigger descriptions** rewritten to steer agents: subscribe + end your turn instead of sleeping/polling.

### Notes

- `time:cron` stays recurring but is now also delivered only to its owning session.
- Legacy subscriptions without a real `subscriptionId` skip targeted deletion (fabricated key path).
- Removes the now-unused `readRunnerId()` helper.

## Testing

- `bun test time-service.reconcile.test.ts` — 26 pass, including new coverage: message in payload, DELETE after fire, past `time:at` immediate fire + cleanup, failed delivery keeps subscription.
- `bunx tsc --noEmit` clean.

### DIFF
--- a/packages/cli/src/runner/services/time-service.reconcile.test.ts
+++ b/packages/cli/src/runner/services/time-service.reconcile.test.ts
@@ -52,6 +52,25 @@ function entry(
     return { sessionId, triggerType, subscriptionId: subscriptionId ?? `${sessionId}-${triggerType}`, runnerId: "runner-test", params };
 }
 
+interface RecordedCall { method: string; url: string; body: string }
+
+/** Install a recording fetch mock. Returns the call log (fires are POSTs, one-shot cleanup are DELETEs). */
+function mockFetch(status = 200): RecordedCall[] {
+    const calls: RecordedCall[] = [];
+    globalThis.fetch = (async (input: string | URL | Request, init?: RequestInit) => {
+        calls.push({
+            method: init?.method ?? "GET",
+            url: typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url,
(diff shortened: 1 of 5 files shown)


Pizzaface:
Added a second commit: **recurring-schedule UI for cron params**. The subscribe form for `time:cron` (or any trigger with a `cron` param) now shows a Daily / Weekly / Monthly picker + native time input instead of a raw cron field. Local time in, UTC cron out (runner evaluates cron in UTC), with day-shift handling across the midnight boundary. "Custom (cron)" option keeps the raw input for steps/ranges/lists. Converters are pure functions in `lib/cron-schedule.ts` with round-trip tests run under UTC, Tokyo, and UTC+14.

PR #612, opened 2026-07-17
feat(runner): track and kill session-spawned processes

Pizzaface:
## What

Sessions spawn OS processes (bash children, MCP stdio servers, dev servers) that were previously untracked — killing a session orphaned anything it started in the background.

- **Process groups**: workers spawn `detached: true` so PGID = worker PID; every descendant inherits the group
- **Kill on exit**: `kill_session`, the SIGKILL escalation, and natural worker exit now signal the whole group (`kill(-pid)`) with fallback to the lone worker
- **Web UI**: new `process` runner service + Processes service panel — lists PID/uptime/mem/command per session with per-process kill (3s polling)
- **TUI**: live dim widget above the editor (`⚙ 2 procs bun run:4123 · vite:4188`), hidden when the session has no children

## Safety

- Panel kill requests are validated against `pgrep -g` group membership — arbitrary system PIDs are rejected, and the worker itself is refused (use Kill Session)
- Restart-in-place (exit 43) intentionally does **not** kill the group, so session background processes survive worker restarts

## Known ceilings (marked with ponytail comments)

- Adopted sessions (worker survived a daemon restart) have no PID handle — panel shows "unavailable"
- Background procs from a pre-restart worker land in an old pgid and aren't reaped on final exit

## Testing

- 10 new tests: ps-line parsing, group kill incl. grandchildren, service list/kill guards, widget formatting
- Full `runner/` suite (368), UI panel suites (141), and repo typecheck pass

### DIFF
--- a/packages/cli/src/extensions/factories.test.ts
+++ b/packages/cli/src/extensions/factories.test.ts
@@ -26,6 +26,7 @@ import { toolSearchExtension } from "./tool-search.js";
 import { ollamaWebToolsExtension } from "./ollama-web-tools.js";
 import { sessionAnalysisExtension } from "./session-analysis.js";
 import { providerRequestLogExtension } from "./provider-request-log.js";
+import { sessionProcessesExtension } from "./session-processes.js";
 
 const CORE_EXTENSIONS: ExtensionFactory[] = [
     providerRequestLogExtension,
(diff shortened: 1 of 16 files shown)


Pizzaface:
!pizza fix the checks

PR #618, opened 2026-07-21
fix(subagent): claude-subscription provider support in subagents

Pizzaface:
## Problem
With the claude-subscription provider (minimalcc-pi) as the only configured Claude provider, the `subagent` tool fails:

1. `model: { provider: "anthropic", id: "claude-haiku-4-5" }` (the form nightshift rules mandate, and what the `haiku`/`sonnet`/`opus` frontmatter aliases parse to) resolves to the credential-less built-in anthropic catalog → **"No API key found for anthropic"**.
2. Even with an explicit `claude-subscription/...` override, the subagent session builds a fresh ModelRegistry (extensions are skipped via `noExtensions: true`), so the extension-registered provider config/key is missing → **"No API key found for claude-subscription"**. (Reproduced live before the fix.)

## Fix
- New `resolveModelSpec` in `engine.ts`: exact match only wins if the provider has credentials; otherwise falls back to Ollama Cloud cache → same-id model under an available provider → the credential-less exact match (so auth errors still name the requested provider).
- `runSingleAgent` now passes the parent session's `ModelRegistry` into `createAgentSession` (guarded so narrow test mocks are not passed), giving subagent sessions access to extension-registered providers and their keys.

## Verification
- `bun test src/extensions/subagent.test.ts src/extensions/subagent-agents.test.ts` — 78 pass, 0 fail (5 new resolveModelSpec tests)
- `tsc --noEmit` clean
- **Live e2e** (`packages/cli/scripts/e2e-claude-sub-subagent.ts`): registers the claude-subscription provider exactly like the extension does, runs a subagent with an `anthropic/claude-haiku-4-5` override → resolves to `claude-subscription/claude-haiku-4-5`, streams a real request, returns `pong`, exit 0.

### DIFF
--- a/packages/cli/scripts/e2e-claude-sub-subagent.ts
+++ b/packages/cli/scripts/e2e-claude-sub-subagent.ts
@@ -1,0 +1,68 @@
+/**
+ * E2E check: subagent engine works with the claude-subscription provider.
+ *
+ * Registers the provider directly on a real ModelRegistry (same call the
+ * minimalcc-pi extension makes), then runs a subagent with an anthropic/*
+ * model override — exercising the same-id provider fallback AND shared
+ * registry auth. Burns one tiny haiku request.
+ *
+ * Run: bun scripts/e2e-claude-sub-subagent.ts
+ */
+import { ModelRegistry, AuthStorage } from "@earendil-works/pi-coding-agent";
+import { join } from "node:path";
+import { homedir } from "node:os";
+import { runSingleAgent } from "../src/extensions/subagent/engine.js";
+import type { AgentConfig } from "../src/extensions/subagent-agents.js";
+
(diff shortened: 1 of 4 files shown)


Pizzaface:
!pizza fix the checks.

Pizzaface:
✅ Checks fixed!

**What happened:**
- The failing `Unit Tests` job timed out waiting for `sandbox-runner` in the browser smoke test.
- The same browser smoke test passed locally.
- Re-ran the failed workflow job; no code change was needed.

**Verification:**
- Unit Tests: ✅
- CodeQL: ✅
- Binary builds: ✅
- E2E install flows: ✅
- Mobile — Android: ✅

All 12 PR checks now pass.

PR #671, opened 2026-08-01
feat(docker): runner container image (ghcr.io/pizzaface/pizzapi-runner)

Pizzaface:
Implements the runner container image per the approved spike (`docs/plans/2026-07-31-runner-container-spike.md`), covering all three slices of Godmother epic `dQR5hwwdI2kSP7AxPsSUq` plus the CLI health probe the spike said was a prerequisite.

`ghcr.io/pizzaface/pizzapi-runner` is a **remote coding machine**, not a daemon wrapper: Debian slim, non-root `pizza` (1000:1000), `tini` as PID 1, built from the existing standalone Linux binaries. It dials out to a relay and exposes no inbound ports.

## What's here

| Area | Change |
|---|---|
| Image | `docker/runner/Dockerfile` — one arch-generic Dockerfile, `linux/amd64` + `linux/arm64` via `TARGETARCH` |
| Entrypoint | `docker/runner/entrypoint.sh` — non-root by default; optional `PUID`/`PGID` remap with privilege drop via `setpriv` |
| Build | `docker/runner/stage-context.ts` — compiles/reuses the standalone binaries and assembles a minimal build context |
| Smoke | `docker/runner/smoke.sh` — 17 checks end-to-end against a live relay |
| Compose | `docker/runner/compose.example.yml` — operator-facing example |
| CLI | `pizza runner status [--json]` — real health probe (alive **and** relay-registered), backs the image `HEALTHCHECK` |
| CLI fix | Stale-lock check now detects PID reuse (found by the smoke test — see below) |
| CI | GHCR publish on release + edge publish on `main` + hermetic Dockerfile check on every PR |
| Docs | `deployment/runner-container.mdx` operator guide + `pizza runner status` in the CLI reference |

## Decisions made

- **Registry:** `ghcr.io/pizzaface/pizzapi-runner`. Release publishes `:<version>` + `:<sha>`, and `:latest` **only** for `latest` releases (never beta/dev). Pushes to `main` publish `:main` + `:<sha>`.
- **UID/GID:** non-root `pizza` 1000 by default (a fresh named volume inherits that ownership automatically). Operators whose bind mounts are owned by another UID run the container as root with `PUID`/`PGID`; the entrypoint remaps the *existing* `pizza` account and drops privileges. This is deliberately not a raw `user:` override — a raw override leaves no `/etc/passwd` entry for the UID (verified: `whoami` fails), which breaks git and shells.
- **Sandbox:** image defaults to `PIZZAPI_SANDBOX=none`. Bubblewrap cannot create user namespaces under Docker's default seccomp profile — verified on Docker 29.5.3, it fails even as root, while `--security-opt seccomp=unconfined` makes it succeed. Rather than ship a silently-degraded sandbox, the default is explicit and the opt-in path is **tested** (smoke check 17) and documented as the isolation trade it actually is.
- **Sandbox assets:** `@anthropic-ai/sandbox-runtime` resolves its `vendor/seccomp/<arch>` helpers relative to its own module dir, which doesn't exist inside a bun-compiled binary. The image stages them at `/usr/lib/node_modules/@anthropic-ai/sandbox-runtime/vendor/seccomp/<arch>/`, one of srt's global fallback locations — so the sandbox works from a standalone binary with **no node/npm in the image**.
- **Healthcheck:** the spike deferred this pending a real probe, so the probe was built. `pizza runner status` exits 0 only when the daemon PID is alive *and* registered with the relay; no relay round-trip, no config/model loading, cheap enough for a 30s interval.

## Bug found and fixed

The smoke test caught a real, reproducible bug outside the image itself. A fresh container resets its PID namespace to 1, so the daemon's process tree gets the same deterministic low PIDs on every start. After `docker kill -9` + recreate on the same volume, the new daemon saw *its own PID number* in the stale lock file and refused to start with a fatal `already running`, self-healing ~2s later only because the supervisor blindly retries. PID liveness + `ps` cmdline matching cannot distinguish "same process still alive" from "brand-new process that reused the PID".

Fixed in `runner-state.ts` by recording the OS-reported process start time alongside the PID (`/proc/<pid>/stat` field 22 on Linux, `ps -o lstart=` elsewhere) and only trusting a lock when both match. Missing or unparseable start times fall back to the previous behaviour, so this is strictly an additional discriminator. Godmother: `Uoe0ItIe`.

## Verification

Smoke run against a live relay (`docker network pizzapi-web_default`, relay at `server:7492`) — **17/17 PASS, 0 FAIL, 0 SKIP**, reproduced across 4 independent runs:

- both arches build; foreign-arch binary runs under emulation
- starts as uid 1000 non-root and reaches Docker `healthy` (~8s) via the real healthcheck
- workspace enforcement: `cwd=/etc` rejected (400), real repo under `/workspace` accepted (200)
- PTY library resolves cleanly; no terminal/PTY errors in the daemon log
- reaches Compose sidecars by DNS name (`server:7492`, `redis:6379`)
- `docker stop` shuts down cleanly in 0–1s (well inside the 30s grace period)
- runner identity survives stop + `rm` + recreate on the same volume
- stale-lock recovery after SIGKILL — healthy again with no fatal error
- no `--privileged`, no host PID namespace, no added capabilities, no Docker socket
- opt-in sandbox verified: with `seccomp=unconfined`, `bwrap --unshare-all` succeeds inside the container

Also: `bun run typecheck` clean; `bun test packages/cli/src` → 2638 pass. The 3 remaining failures are pre-existing and unrelated — `origin/main` at equal path depth fails 4 in the same suite on this machine (models-command + computePackageIdentity, environment/pollution sensitive). `bun run --cwd packages/docs build` → exit 0, 49 pages.

Image size: 341MB arm64, 318MB amd64.

## Notes for review

- CI's image job is **hermetic** (single-arch, `--load`, no push) and asserts the CLI runs, uid is 1000, `pizza runner status` fails closed with no daemon, required tools are on PATH, seccomp assets are staged, and env defaults are right. The relay-backed smoke stays in `docker/runner/smoke.sh` because CI has no provisioned relay or API key.
- The release job **reuses the release-stamped binaries** from the `build-binary` matrix rather than recompiling, so the image ships the same bits as the npm release.
- A QEMU setup step was added to the new release job; the existing `docker-ui` job builds multi-arch without one, which may be a latent gap there (left untouched — out of scope).

Closes `ncNx1zxm`, `iQ6y7fgP`, `oOwIFNPU`, `pcBd0QmU`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,8 +309,8 @@ jobs:
           build-args: |
             VERSION=ci
             GIT_SHA=${{ github.sha }}
-          cache-from: type=gha,scope=runner-image
-          cache-to: type=gha,mode=max,scope=runner-image
+          cache-from: type=gha,scope=runner-image-ci
+          cache-to: type=gha,mode=max,scope=runner-image-ci
 
       - name: Assert — CLI works and image is non-root
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,6 +257,99 @@ jobs:
           retention-days: 7
 
   # ───────────────────────────────────────────────────────────────
+  # Runner image — hermetic Dockerfile check (linux/amd64 only, --load,
+  # no push). Catches Dockerfile/entrypoint breakage on every PR without
+  # needing a relay. The full relay-backed smoke (registration, spawn,
+  # persistence, stale-lock recovery, sandbox opt-in) lives in
+  # docker/runner/smoke.sh and is run locally against a live relay,
+  # because CI has no provisioned relay/API key.
+  # ───────────────────────────────────────────────────────────────
+  runner-image:
+    name: Runner Image — hermetic check
+    needs: test
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: "1.3.10"
+
+      - name: Cache Bun artifacts
+        uses: actions/cache@v5
+        with:
+          path: |
+            ~/.bun/install/cache
+            node_modules
+          key: ${{ runner.os }}-${{ runner.arch }}-bun-${{ hashFiles('bun.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ runner.arch }}-bun-
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Build packages
+        run: bun run build
+
+      - name: Stage Docker build context (linux-x64 only)
+        run: bun docker/runner/stage-context.ts --targets linux-x64
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Build runner image (amd64, load only)
+        uses: docker/build-push-action@v7
+        with:
+          context: packages/cli/dist/runner-image-context
+          file: docker/runner/Dockerfile
+          platforms: linux/amd64
+          load: true
+          tags: pizzapi/runner-image:ci
+          build-args: |
+            VERSION=ci
+            GIT_SHA=${{ github.sha }}
+          cache-from: type=gha,scope=runner-image
+          cache-to: type=gha,mode=max,scope=runner-image
+
+      - name: Assert — CLI works and image is non-root
+        run: |
+          set -euo pipefail
+          IMG=pizzapi/runner-image:ci
+
+          docker run --rm --entrypoint pizza "$IMG" --version
+
+          docker run --rm --entrypoint sh "$IMG" -c '
+            set -eu
+            [ "$(id -u)" = "1000" ] || { echo "expected uid 1000, got $(id -u)" >&2; exit 1; }
+            for c in bwrap socat rg git ssh tini; do
+              command -v "$c" >/dev/null || { echo "missing $c on PATH" >&2; exit 1; }
+            done
+            SECCOMP_DIR=/usr/lib/node_modules/@anthropic-ai/sandbox-runtime/vendor/seccomp/x64
+            [ -x "$SECCOMP_DIR/apply-seccomp" ] || { echo "apply-seccomp missing/not executable" >&2; exit 1; }
+            [ -f "$SECCOMP_DIR/unix-block.bpf" ] || { echo "unix-block.bpf missing" >&2; exit 1; }
+            [ "$PIZZAPI_SANDBOX" = "none" ] || { echo "PIZZAPI_SANDBOX default wrong: $PIZZAPI_SANDBOX" >&2; exit 1; }
+            [ "$PIZZAPI_WORKSPACE_ROOTS" = "/workspace" ] || { echo "PIZZAPI_WORKSPACE_ROOTS default wrong: $PIZZAPI_WORKSPACE_ROOTS" >&2; exit 1; }
+            echo "✅ uid, tools, seccomp assets, and env defaults all check out"
+          '
+
+      - name: Assert — healthcheck command fails closed without a daemon
+        run: |
+          set -euo pipefail
+          IMG=pizzapi/runner-image:ci
+
+          set +e
+          docker run --rm --entrypoint pizza "$IMG" runner status
+          CODE=$?
+          set -e
+
+          if [ "$CODE" -ne 1 ]; then
+            echo "::error::expected 'pizza runner status' to exit 1 with no daemon running, got $CODE"
+            exit 1
+          fi
+          echo "✅ pizza runner status correctly exits 1 without a daemon (HEALTHCHECK fails closed)"
+
+  # ───────────────────────────────────────────────────────────────
   # E2E: install flow (`pizzapi web` + register + runner) per distribution
   # ───────────────────────────────────────────────────────────────
   e2e-install-flow:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -459,6 +459,94 @@ jobs:
           cache-from: type=gha,scope=ui-image-release
           cache-to: type=gha,mode=max,scope=ui-image-release
 
+  # ─── Build + push versioned Docker runner image ───────────────
+  # Reuses the release-stamped binaries from build-binary instead of
+  # recompiling, so the image contains the exact same bits as the npm
+  # release. needs: publish (which needs: build-binary) guarantees the
+  # binary-linux-* artifacts already exist by the time this job starts.
+  docker-runner:
+    name: Docker Runner — ${{ needs.compute-version.outputs.version }}
+    needs: [compute-version, publish]
+    if: inputs.action != 'pull' && inputs.dry-run != true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: "1.3.10"
+
+      - name: Install dependencies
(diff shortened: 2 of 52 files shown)


Pizzaface:
## Update: container auth reworked

Pushed a second body of work onto this PR. The original slices are unchanged; this replaces the container's credential story, which was the weakest part of the first pass (you had to paste a long-lived relay key into Compose, and getting provider credentials in meant hand-copying `auth.json` into the data volume).

### Relay auth: device pairing instead of a pasted key

A container with no credential now pairs itself:

```
no credential in volume or env, relay URL known
  → POST /api/setup-claim { relayUrl, label }
  → approval URL + QR printed to docker logs (re-printed ~every 60s)
  → you approve in an authenticated browser
  → relay mints a dedicated 1-year key named runner-<label>
  → persisted 0600 into the mounted volume; daemon continues, no restart
```

No secret in Compose, none in `docker inspect`, and each container gets its own key that shows up by name in **Settings → API Keys** with a delete button. `PIZZAPI_PAIRING=0` restores the previous fail-fast behaviour for operators who prefer injecting a key.

### Provider auth: secrets files instead of copying `auth.json`

| Variable | Behaviour |
|---|---|
| `<NAME>_FILE` | Populates `<NAME>` from the trimmed file contents. Scoped to credential-shaped names (`_KEY` / `_TOKEN` suffixes, which covers `_API_KEY`); an already-set var wins |
| `PIZZAPI_AUTH_FILE` | Seeds `auth.json` (0600) **only when absent** — never clobbers, since OAuth refresh writes back to it |

The scoping is deliberate: a blanket `*_FILE` rule would happily load something like `NTFY_AUTH_FILE=/var/lib/ntfy/auth.db` into an env var as binary garbage.

### Three bugs found and fixed along the way

**1. The approval page could burn the key it was about to grant.** The non-consuming claim read was first added at `GET /api/setup-claim/:token/info`. But every deployed relay resolves that route's token as `pathname.slice(prefix).split("/")[0]`, so the nested path doesn't 404 on an older server — it reaches the **one-shot redeem handler with a valid token** and silently consumes the approved claim. The UI ships as a separately versioned image (`docker/compose.yml` pulls `pizzapi-ui:latest` while building the server from source), so newer-UI-against-older-server is a normal deployment. Moved to `/api/setup-claim-info/:token`, which old relays simply 404. The regression test asserts *both* halves — that the nested path is genuinely consuming, and that the dedicated prefix isn't.

**2. A stale API key returned 500 instead of 401 on every session-authenticated route.** better-auth's api-key plugin hooks session resolution, so a malformed `x-api-key` makes `getSession()` throw before our explicit validation runs. Reproduced live: the same bad key gives 401 on `/api/runners/spawn` (which reaches `validateApiKey` directly) but 500 on `/api/runners`, `/api/sessions`, and the approve route. `routes/tunnel-ws.ts` and `ws/namespaces/auth.ts` already guarded this; `middleware.ts` was the outlier. Pre-existing on `main`, and it sends anyone with an expired key off debugging the server instead of their credential.

Note the existing test `validateApiKey rejects invalid keys` passed the entire time — because it calls the function directly and never goes through `getSession`. The new tests go through the route layer; verified they fail without the fix and pass with it.

**3. `smoke.sh` could die in its own cleanup trap.** On macOS's bash 3.2, `"${arr[@]}"` on an empty array is fatal under `set -u`. Exiting at a preflight FATAL — before any result was recorded — made the trap throw "unbound variable" and swallow the actual error, which is exactly what a first-time user hits.

### Verification

- **`docker/runner/smoke.sh`: 30 PASS / 1 SKIP / 0 FAIL** against the live relay, including container pairing end to end (approval URL appears → approved → healthy → `config.json` 0600 with the key → restart does **not** re-pair), `PIZZAPI_PAIRING=0` failing fast, `PIZZAPI_API_KEY_FILE` registering with the key absent from `docker inspect`, `PIZZAPI_AUTH_FILE` seeding without clobbering, and the loud bad-key rejection.
- The one SKIP is honest: the live relay runs a released build that 404s on `/api/setup-claim-info`, so label round-trip can't be exercised there. Verified separately against a server built from this branch: label persisted and returned, `/info` after approval returned `approved` **without** the key and without redeeming, the CLI poll then redeemed normally, and the minted key was named `runner-docker-demo-runner` with a 1-year TTL.
- `bun run typecheck` clean; `bun scripts/test-isolated.ts packages/server/src` → 73 passed; `bun test packages/cli/src` → 2670 passed (3 pre-existing failures, `origin/main` fails 4 in the same suite on this machine); docs build → exit 0.

### What this does *not* fix

Pairing scopes and makes the credential revocable — it does **not** isolate sessions from it. The daemon still passes its relay key into every worker's environment (`packages/cli/src/runner/session-spawner.ts:203`), because sessions connect to the relay themselves, so any agent running in the container can read it. Pairing downgrades the blast radius from your master key to a scoped, revocable, per-container one. The docs say this plainly rather than implying pairing makes it safe. Real containment would mean per-session credentials — a separate piece of work.

Pizzaface:
## Added: `pizza runner pair [--force]`

Closes the last rough edge in the container auth story. Re-pairing after you revoke a container's key previously meant deleting a file inside a Docker volume and restarting; now it's a command.

```
pizza runner pair            # refuses if a credential already resolves — exit 1
pizza runner pair --force    # pairs and replaces the existing credential — exit 0
```

Runs the same headless flow the supervisor uses (approval URL + QR to stdout, polls, writes `0600` into the agent dir), sends the same `runner-<name>` label, and replaces **only** the credential fields in `config.json`.

### The guard rail that justifies the command

`PIZZAPI_API_KEY` in the environment **wins** over the stored credential. That's exactly how the old docs told people to configure a container — so without a guard, an operator upgrading to pairing could run `pair --force`, see "paired successfully", and still be broken by a leftover env var, with the freshly minted key silently shadowed.

So the command **refuses** when any of `PIZZAPI_RUNNER_API_KEY` / `PIZZAPI_API_KEY` / `PIZZAPI_API_TOKEN` is set, naming the offending variable, and checks *before* burning a claim. Warn-and-continue was rejected deliberately: operators scan for the success line and won't read a warning printed above it. `--force` is the single conscious escape hatch, and the warning still prints.

It also prints the next step when a daemon is already running (`pizza runner stop` / `docker restart <container>`) rather than auto-restarting anything.

### Bug found and fixed: re-pairing was broken by a `ws://` round-trip

Pairing persists `relayUrl` in `ws(s)://` form because that's what the daemon's socket.io connection wants — but the `setup-claim` REST call needs `http(s)://`. So the **first** pair worked (URL came from env in `http://` form) and every subsequent one, which reads it back from config, died with `protocol must be http:, https: or s3:`.

This was already latently reachable from the existing auto-pair path (key revoked, `relayUrl` left set), not just the new command — a bug that only reproduces on the *second* run, which is precisely the shape a quick manual test misses. Fixed by lifting `daemon.ts`'s local `normaliseRelayUrl` into shared `relay-url.ts#toHttpRelayUrl` and using it at all three call sites, so there aren't competing spellings of the same URL logic left to diverge. Regression test pins the config-sourced `ws://` case specifically.

### Verification

- `bun run typecheck` → exit 0; `bun test packages/cli/src` → **2695 pass** (3 pre-existing failures unchanged)
- `bun run --cwd packages/docs build` → exit 0
- `./docker/runner/smoke.sh` → full run passes, plus a new check that `pair` without `--force` refuses on an already-paired container **and leaves the existing credential untouched**
- Live runs against the relay with an isolated HOME: refusal without `--force`; refusal for each of the three shadowing env vars, config untouched; `--force` writing a genuinely different key at `0600` with unrelated config keys surviving; the key never appearing in output (grep count 0); and the daemon-running advice line firing against a process that actually matches the liveness pattern

PR #695, opened 2026-08-11
fix(goal): gate the goal loop on run boundaries, not every LLM turn

Pizzaface:
## The bug

`/goal` hooked `turn_end`, which pi fires after **every LLM round-trip** — including `stopReason: "toolUse"` turns where the agent is mid-run and about to call more tools on its own. Only `aborted`/`error` were filtered, so every tool-call turn was treated as a decision point.

| Symptom | Cause |
|---|---|
| "Work toward this goal" injected after every tool call | steer sent on mid-run turns while below `min-turns` (default 10) or between throttled evals |
| `--max-turns 20` exhausted by one complex task | `turnCount` counted LLM calls, not agent responses |
| Evaluator judging half-finished work, blocking the loop | LLM judge ran mid-run and was awaited inside the agent loop |

No test covered `stopReason: "toolUse"` — 1,700 lines of goal tests, all `"stop"`.

## The fix

Only `stop`/`length` turns end a run. Mid-run turns accrue token/cost spend and still enforce budgets (a runaway tool loop can't overspend) but never steer, evaluate, or advance `turnCount`. `recordTurnSpend` is split from a new `recordCompletedRun`, and state persists at run end so a crash no longer loses accrued spend.

## Also fixed

- **The system prompt lied.** It told the agent "the session stops and configured stop hooks run" — `/goal` never stops the session and no stop hooks exist. Replaced with what actually happens, plus a note to end runs with evidence rather than assertion.
- **Expensive-by-default evaluator.** Default was the session's own model (e.g. Opus), justified as sharing the prompt cache — but the evaluator prompt shares no prefix with the session, so nothing cached; it just billed a 60k-char transcript to the priciest model available. Now defaults to the cheapest authenticated text model, and a configured-but-unauthenticated model falls back instead of silently disabling evaluation.
- **Self-report loop.** The evaluator prompt now demands tool-output evidence (exit status, test results, file contents) instead of accepting an assistant's claim that the goal is done.
- **Uneven cadence.** `turnCount % rate` spaced evaluations unevenly when `min-turns` wasn't a multiple of the rate; now runs-since-last-eval. The first eligible run is still always evaluated.
- **Defaults retuned for run semantics:** `min-turns` 10 → 1, `every` 3 → 1. A run is already substantial work, and skipping a check costs a whole extra agent run — far more than one small judge call.
- **Guidance delivered twice.** It went into both the system prompt (`before_agent_start`) and the steer message, and was never cleared from the prompt, so it repeated every turn. Dropped the handler; the steer message carries it.
- **`--max-turns 0`** was accepted, setting a budget exhausted before the first run.
- **Verdict JSON** used a non-greedy `{...}` match that truncated on braces inside `reason`, falling through to fragile free-text regex.
- **Config read 2–3× per `turn_end`** → loaded once.
- **Three copy-pasted budget/steer blocks** → `enforceBudget` + `steerContinuation`, preserving the rule that only `not_met` auto-continues (an `uncertain` verdict must not spin the session forever).

## Surface updates

Web UI settings panel (`FastModelSettings`) defaults and labels now match the CLI and say "runs" instead of "turns"; `configuration.mdx` and `slash-commands.mdx` updated.

## Testing

`bun test packages/cli/src/extensions/goal` — 101 pass, including new regression coverage:

- mid-run `toolUse` turns accrue spend but never steer, evaluate, or count a run
- a mid-run tool loop still stops the goal when a token budget is exhausted
- an `uncertain` verdict does not auto-continue
- cadence measured from the last evaluation
- `--min-turns` deferral and the new first-run-evaluates default
- guidance rides the steer message only; no `before_agent_start` registered
- zero/negative budget rejection, JSON reasons containing braces

`bun run typecheck` clean. Full `bun test packages/cli/src`: 2810 pass / 3 fail — the same 3 failures are present on a clean `origin/main` (cross-file pollution in `models-command` / `identity`, unrelated). UI component suite likewise unchanged from baseline.


Pizzaface:
Follow-up commit: added `features/goals.mdx` (Session Goals) and aligned the remaining "turn" wording.

The audit found `/goal` was documented as three table rows plus a config block — the surprising parts were nowhere: that it starts new agent runs on its own, that it never terminates the session, that budgets count runs, and that the LLM evaluator receives the session transcript. That last one existed only as a comment in `transcript.ts`, so it's now a proper privacy warning with local-only alternatives (`--evaluator keyword`, or a local `evaluatorModel`).

Also renamed the user-visible counters now that `turnCount` counts runs: the web UI header badge and TUI status line read `run N/M`, and `/goal status` reports `Runs:`. Leaving the flag itself as `--max-turns` (renaming it would break existing usage); the docs and tooltips state the unit explicitly instead.

Three claims I checked against the code rather than assuming: the settings section is labelled **Fast Model**, not "Goal Evaluator"; the badge really is a session-header element; and `TRANSCRIPT_MAX_CHARS` is 60k. Docs build clean, 109 goal/badge/settings tests pass, typecheck clean.

Pizzaface:
Follow-up commit: the evaluator now runs on the session's model **by reusing the session's cached context**, and the Fast Model picker is gone.

Making the cache real needed a change to the call *shape*, not just the model. The old evaluator sent a standalone request (one user message, no system prompt, no tools), which shares no prefix with the session — so pointing it at the session model would have paid full input price on the priciest model available, plus a cache-*write* premium (pi-ai auto-applies `cache_control` to the last block of every request) for an entry nothing ever reads.

The request is now `[session system prompt + session tools + session messages + judge question]` — the shape of a normal follow-up turn. The prefix hits the cache the session already paid for, and only the appended question is billed as new input. Two bonuses: the judge sees the real conversation instead of a 60k-char truncation, and evaluation no longer ships session text to a provider the session wasn't already using (the privacy caveat in the docs now applies only to a pinned `evaluatorModel`).

**Verified, not assumed** — a hit depends on reproducing the prefix byte for byte:

- The prefix is captured from pi's own `context` event + `ctx.getSystemPrompt()` + `pi.getAllTools()/getActiveTools()`, not reconstructed from guesses.
- `context` fires *before* pi's `convertToLlm` step, so custom agent messages (our background-bash execution messages) are filtered out. That approximates the conversion — it doesn't claim to reproduce it.
- Every evaluation checks the provider's reported `cacheRead`. Zero means the assumption failed: the session stops reusing context and falls back to a standalone prompt on the cheapest authenticated model, logged once. **A mismatch costs one evaluation, not a session of full-price transcripts.**

`goal.evaluatorModel` survives as an advanced escape hatch (pin a local/private judge); setting it opts out of cache reuse. Settings section renamed **Fast Model → Goal Evaluator**, keeping max-tokens and cadence — and saving from it preserves a pinned model it no longer edits (regression-tested).

Tests: prefix fidelity, no-transcript-resend, cache reporting, the viability tracker, `convertToLlm` filtering, plus integration coverage for both the cache path and the post-miss fallback. Full CLI suite 162/162 files, UI suite 1270 tests, typecheck and docs build all clean.

One caveat worth stating plainly: even with a hit, a very large conversation on an expensive model can cost more per evaluation than the old cheap-model call on a truncated transcript (cache reads are ~0.1× input, not free). The cache-read check protects against the catastrophic case, not against Opus-sized prefixes. If evaluator spend looks high in the Usage dashboard, `--every N` throttles it.

PR #735, opened 2026-08-16
fix(schedules): resume the scheduling session transcript when a schedule wakes

Pizzaface:
## Problem

When a schedule fires and the session that created it is gone, the time service spawns a replacement session with the schedule instruction as its prompt (#734). That session starts **empty** — it has no memory of the conversation that scheduled the work.

Observed live: a `time:timer_fired` test woke a fresh session which confidently "confirmed" the timer using only the trigger payload. Asked what we had discussed before, it had nothing — the scheduling conversation was in a different session's transcript.

## Fix

Carry the transcript path the same way `_cwd` is already carried:

- `subscribe_trigger` stamps `_resumePath` (from `PIZZAPI_SESSION_FILE`, which the worker keeps current across `/new` and resume) into `time:*` subscription params.
- `#spawnReplacementSession` passes it as `resumePath` on `POST /api/runners/spawn`.
- Cron re-own carries `_resumePath` into the migrated subscription.

Everything downstream already existed and is unchanged: the spawn route forwards `resumePath`, the daemon sets `PIZZAPI_WORKER_RESUME_PATH`, and `initial-prompt.ts` calls `switchSession()` *before* sending the trigger prompt.

`SessionManager.open()` appends to the same file, so a repeatedly-migrated cron keeps resuming one continuous thread rather than fanning out into new transcripts.

## Notes

- A stale/deleted transcript is not fatal: the worker logs a warning and starts fresh — today's behavior. No daemon-side `stat` added.
- Live sessions are untouched; this only affects the "owner is gone" replacement path.

## Testing

- `bun test packages/cli/src/runner/services packages/cli/src/extensions/triggers` — 269 pass, 0 fail (3 new assertions: spawn body `resumePath` for one-shot and cron, `_resumePath` in re-own params).
- `bunx tsc --noEmit -p packages/cli/tsconfig.json` — clean.
- Requires a daemon restart to take effect on this runner.

### DIFF
--- a/packages/cli/src/extensions/triggers/extension.ts
+++ b/packages/cli/src/extensions/triggers/extension.ts
@@ -790,9 +790,18 @@ export const triggersExtension: ExtensionFactory = (pi) => {
 
             // Schedules can outlive the session that created them: capture the
             // workspace so the time service can spawn a replacement session in
-            // the right cwd if this session is gone when the schedule fires.
+            // the right cwd if this session is gone when the schedule fires,
+            // plus this session's transcript so that replacement RESUMES this
+            // conversation instead of waking with no memory of it.
+            // PIZZAPI_SESSION_FILE is kept current by the worker across /new
+            // and resume, so it always names the live transcript.
             if (params.triggerType.startsWith("time:")) {
-                subParams = { ...(subParams ?? {}), _cwd: process.cwd() };
+                const sessionFile = process.env.PIZZAPI_SESSION_FILE?.trim();
+                subParams = {
+                    ...(subParams ?? {}),
+                    _cwd: process.cwd(),
+                    ...(sessionFile ? { _resumePath: sessionFile } : {}),
+                };
             }
 
             // Coerce filter values
--- a/packages/cli/src/runner/daemon.ts
+++ b/packages/cli/src/runner/daemon.ts
@@ -34,6 +34,7 @@ import { TunnelClient } from "@pizzapi/tunnel";
 import { loadGlobalConfig, saveGlobalConfig, defaultAgentDir, expandHome, loadConfig } from "../config.js";
 import type { PizzaPiConfig } from "../config.js";
 import { findSessionPathById } from "./session-list-cache.js";
+import { lookupTranscriptLink } from "./session-transcript-links.js";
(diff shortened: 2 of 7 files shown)


Pizzaface:
Second commit added after finding the real root cause live on this runner.

### The bug this actually fixes

A relay session id and a pi transcript id are different identifiers. `wakeOfflineSession` sends `resumeId: <relaySessionId>`, but `findSessionPathById` matches the **pi** id from the `.jsonl` filename — so the wake resume never resolved, and a plain respawn sent no resume info at all. Both restarted the session on an empty transcript.

Caught in the runner log — same relay session, two transcripts across a daemon restart:

```
16:13:40 [daemon] spawning headless worker for session 08ce0b86…
16:13:41 [daemon] session 08ce0b86 reported transcript …_01a00b59-….jsonl
         — daemon restarted at 16:16:24 —
16:18:00 [daemon] spawning headless worker for session 08ce0b86…
16:18:02 [daemon] session 08ce0b86 reported transcript …_01a00b5d-….jsonl   ← new, empty
```

The daemon already learns the pairing from the worker's `session_metadata` IPC, but kept it in memory only. It is now persisted to `~/.pizzapi/session-transcript-links.json` and consulted first when resolving a resume, which fixes the wake path, the plain respawn path, and the schedule-replacement path together.

### End-to-end verification on a live runner

Spawned a disposable child session, had it print a secret and arm a 120s timer, let its worker exit, then let the relay wake it:

```
16:58:05  child reports transcript …_01a00b82.jsonl        → link recorded automatically
16:59:00  child worker exits                                → session offline
17:00:25  daemon: resuming session a2500a8c from its recorded transcript …_01a00b82.jsonl
17:00:37  worker: switched to session …_01a00b82.jsonl
17:01:25  timer delivered → child answers "SECRET=PINEAPPLE-7719" from before it died
```

Tests: 5 new unit tests for the link store (record/overwrite/deleted-file/corrupt-store/pruning), plus the 269 existing runner-service and trigger tests. `tsc --noEmit` clean.

### Known cosmetic side effect (pre-existing)

A resumed worker creates its own transcript at boot and *then* `switchSession()`s to the recorded one, so each wake leaves one empty `.jsonl` behind (`…_01a00b84.jsonl` above). That is existing `resumePath` behavior, not new here — worth a separate cleanup if the noise matters.

PR #747, opened 2026-08-17
fix: eliminate 'still connecting' banner stalls + faster connect path

Pizzaface:
## Why

The composer banner `Still connecting to this session — the runner may be offline. Your draft is preserved.` was showing far too often. Three parallel research subagents (client / relay server / runner) mapped every path that can hold `viewerStatus` at `Connecting…` / `Loading session (x of y)…` past the 10s banner threshold, plus live prod-log evidence. This PR fixes every actionable cause and speeds up the happy path.

## Stall fixes

- **Relay no longer fatals on transient Redis errors.** `SocketClosedUnexpectedlyError` / `ConnectionTimeoutError` carry no errno `code`, so `handleProcessError` classified them fatal and shut the whole relay down (disconnecting every viewer and runner) over blips node-redis auto-recovers from. Observed live during a Redis container restart.
- **Stale mid-flight chunk streams expire (15s).** A runner that hung without disconnecting left `pendingChunkedStates` set forever; the `chunkedPending` gate then made every hydration skip the snapshot cache and wait on a runner signal that never came — unrecoverable even across client retries. TTL kept below the client's retry window so the second retry passes the gate.
- **Reconnects no longer re-arm the banner on hydrated sessions.** The socket.io `connect` handler force-wrote `Connecting…` outside the phase machine; the banner keys off the status *string*, so every mobile background / WiFi blip re-armed it on sessions that were visibly fine. Highest-frequency cause.
- **Watchdogs stay armed during chunked transfers.** The chunk header cleared `awaitingSnapshot`, disarming both the hydration-stall and stale-connection watchdogs — a stream stalling mid-way froze `Loading session (x of y)…` with no retry, ever. Chunk arrivals still count as progress.
- **`resync` "already current" now compares against the true seq counter.** `broadcastSessionEventToViewers` advances the seq counter without appending to the Redis cache list, so the cache tail lags — a genuinely-behind viewer matching the stale tail was answered with silence forever.
- **Runner recovery signal is verified.** It was fire-and-forget into a possibly-empty room; now uses `emitToRelaySessionVerified` and tells the viewer honestly when no runner socket is present instead of spinning through the full watchdog window.
- **SQLite snapshot tier unlocks on stale heartbeats.** A hung runner kept `isActive=true` for 2+ minutes (until the orphan sweep), blocking the persisted tier for exactly the sessions that need it. Now gated on a fresh heartbeat (45s = 4 missed beats).

## Speed

- WebSocket-first transport on hub/viewer sockets (polling fallback kept) — the default polling→upgrade handshake cost 2–3 extra RTTs on every connect *and* reconnect.
- Immediate `socket.connect()` on tab-foreground / network-online, bypassing socket.io's up-to-30s backoff.
- First hydration retry at 3.5s (later retries stay at 8s).
- `journal_mode=WAL` + `synchronous=NORMAL` + `busy_timeout` on auth.db — the DELETE journal made every multi-MB `relay_session_state` write pay journal create/fsync/delete on the event loop (measured as ping-timeout churn in prod on the VirtioFS bind mount).
- WAL-speed writes exposed a real nondeterminism: `runner_recent_folder` pruning tie-broke on colliding ISO-ms timestamps; added an `id` tie-break (caught by its own test).

## Deferred (captured in Godmother)

- Runner-side chunk delivery acks, connected-signal debounce, message-size caching (`ebvzX6DF`)
- Partial chunk buffers for mid-stream joiners, cluster-shared chunk-in-flight state, sweep-loop SQLite yields (`0DT6yqSo`)

## Verification

- `bun run typecheck` clean
- `bun scripts/test-isolated.ts packages/server/src` — 80/80 files
- `cd packages/ui && bun test src` — 1480 pass
- New regression tests: transient-redis classification, stale chunk expiry, true-seq resync check, `runnerLooksLive`, verified recovery signal

### DIFF
--- a/packages/server/src/auth.ts
+++ b/packages/server/src/auth.ts
@@ -411,6 +411,7 @@ export function createAuthContext(config: AuthConfig = {}): AuthContext {
     const trustedOrigins = [...baseOrigins, ...extraOrigins, ...mobileOrigins];
 
     const sqliteDb = new Database(dbPath);
+    applySqlitePerfPragmas(sqliteDb);
     const dialect = new BunSqliteDialect({ database: sqliteDb });
     const db = new Kysely<DB>({ dialect });
     const auth = createBetterAuth({
@@ -502,9 +503,28 @@ export function getDisableSignupAfterFirstUser(): boolean {
 
 export type Auth = AuthInstance;
 
+/**
+ * WAL + NORMAL sync: bun:sqlite writes are synchronous on the event loop, and
+ * the default DELETE journal makes every multi-MB relay_session_state write
+ * pay journal create/fsync/delete — measured as ping-timeout churn in prod
+ * (especially on VirtioFS bind mounts). WAL appends instead, cutting write
+ * stalls dramatically. Single-process access, so WAL's shm requirements are
+ * trivial. Best-effort: a filesystem that refuses WAL keeps the default.
+ */
+function applySqlitePerfPragmas(db: Database): void {
+    try {
+        db.run("PRAGMA journal_mode = WAL");
+        db.run("PRAGMA synchronous = NORMAL");
+        db.run("PRAGMA busy_timeout = 5000");
+    } catch (err) {
+        console.warn("[auth-db] Could not apply SQLite perf pragmas (keeping defaults):", err);
+    }
+}
+
 /** Create a standalone Kysely instance for ad-hoc test setup. */
 export function createTestDatabase(dbPath: string): Kysely<DB> {
     const sqliteDb = new Database(dbPath);
+    applySqlitePerfPragmas(sqliteDb);
     return new Kysely<DB>({ dialect: new BunSqliteDialect({ database: sqliteDb }) });
 }
 
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1,4 +1,5 @@
 import { createAuthContext, runWithAuthContext } from "./auth.js";
+import { isTransientRedisError } from "./transient-redis-error.js";
 import { handleFetch } from "./handler.js";
 import {
     getEphemeralSweepIntervalMs,
@@ -47,6 +48,18 @@ function handleProcessError(err: Error, origin: string): void {
     // level and continue rather than triggering a full shutdown.
     if (code === "ECONNRESET" || code === "ERR_STREAM_DESTROYED" || code === "ENOTCONN") {
         processLog.warn(`Caught transient network error (${code}) via ${origin} — logging and continuing:`, err.message);
+        return;
+    }
+
+    // node-redis transient connection errors carry NO ErrnoException code —
+    // they are plain Error subclasses identified by name. The client
+    // auto-reconnects on its own, so a dropped/slow Redis (restart, redeploy,
+    // event-loop stall past the connect timeout) must not shut the relay down;
+    // that would disconnect every viewer and runner over a blip Redis itself
+    // recovers from. Seen live: SocketClosedUnexpectedlyError during a Redis
+    // container restart fataled the whole server.
+    if (isTransientRedisError(err)) {
+        processLog.warn(`Caught transient Redis error (${err.name}) via ${origin} — logging and continuing:`, err.message);
         return;
     }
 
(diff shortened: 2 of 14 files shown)


Pizzaface:
## Adversarial review — 3 rounds (gpt-5.6-sol), all findings triaged

**Round 1 (all fixed in 94769e74):** stale chunk streams are flagged not deleted + recovery requested instead of suppressed; `emitToRelaySessionVerified` gains a node-local room fallback for degraded Redis; chunked headers reset the hydration watchdog.

**Round 2 (all fixed in 76542528):** stale threshold lowered to 10s so the client's ~12s final retry actually exercises the bypass; `activateSession` gains a monotonic activation token re-checked after every await; re-click retry gates on `hydrated` instead of `awaitingSnapshot`.

**Round 3 (fixed in 82c2cbaa):**
- ✅ Superseded activations no longer remove the viewer from a room the current activation shares (rooms aren't reference-counted).
- ✅ When verified recovery reaches no runner socket, cursorless viewers retry the persisted SQLite tier before erroring (the 45s `runnerLooksLive` window otherwise outlasted the client's ~20s retry budget), with the late callback bound to the activation token.
- ⏭️ *Stale-checkpoint "permanently stale" claim — pushed back:* when the cache serves a pre-stream checkpoint we fire a **verified** recovery signal; a hung-but-connected runner processes it once it unhangs (buffered in-order delivery), and a disconnected runner re-emits `session_active` unconditionally on `registered`. Both paths refresh the viewer without broadcasting a multi-MB assembled frame (the thing chunking exists to avoid).
- ⏭️ *Pre-existing, out of scope:* non-atomic runner-secret first claim and the recent-folder upsert missing a unique index — captured as Godmother `EqNrZtr1`.

Pizzaface:
## Review rounds 4–5

**Round 4 (fixed in 81769ab7):** tri-state `emitToRelaySessionChecked` — Redis-degraded lookups return `"unknown"` and are never treated as proof of offline; the persisted fallback is seq-guarded (before + after its lookups) and one-shot/activation-scoped, shared by both the in-activation flush and the client `connected` echo so timing can't fork behavior.

**Round 5 (fixed in de889e10):**
- ✅ Cluster-aware presence: `adapter.sockets()` is local-only under the Redis adapter — switched to `fetchSockets()`. This also strengthens the *pre-existing* `emitToRelaySessionVerified` used by trigger/OAuth forwarding, which had the same blind spot.
- ✅ Pre-send recheck: the fallback re-signals recovery and requires the room to *still* be empty immediately before sending persisted state, so a runner that joined-but-hadn't-broadcast gets the signal instead of being overwritten.
- 📝 *Presence ≠ delivery* (P2): acknowledged in code — callers needing hard delivery proof already have `emitToRelaySessionAwaitingAck`; the viewer path additionally self-heals via the runner's unconditional snapshot-on-register. Pre-existing semantics, unchanged.

**Convergence:** rounds 1→5 went from structural stalls → timing alignment → activation races → ms-scale TOCTOU windows whose ceilings are now documented inline (`ponytail:` markers). Remaining known items are pre-existing and tracked in Godmother (`EqNrZtr1`, `ebvzX6DF`, `0DT6yqSo`). Gates green after every round: full typecheck, 80/80 isolated server files, 1480 UI tests.

Pizzaface:
## Review rounds 6–7 (final)

**Round 6 (1eb53c91):** Redis outage can no longer hang the presence check — local runners resolve via the synchronous node-local room map (no Redis dependency at all), and the cluster `fetchSockets()` is independently bounded to 3s because the Redis adapter awaits `serverCount()` *before* arming its own request timeout. Lookup failure maps to `"unknown"`, never `"empty"`.

**Round 7 (4a52e2d7):** timed-out lookups no longer accumulate in node-redis's offline queue — cluster presence lookups are coalesced per room, so outage-time retries reuse the single pending command instead of replaying a request storm on reconnect.

**Closing the loop.** Seven rounds, 17 findings fixed, 1 pushed back with reasoning, 3 pre-existing items tracked in Godmother. The trajectory: structural hydration stalls → client/server timer alignment → activation races → ms-scale TOCTOU ceilings (documented inline) → Redis-adapter internals. Each remaining theoretical window is either self-healing via the runner's unconditional snapshot-on-register or explicitly marked with its upgrade path. Declaring convergence; further hardening belongs to the tracked follow-ups (`ebvzX6DF`, `0DT6yqSo`, `EqNrZtr1`).

Gates after every round: `bun run typecheck` clean · `bun scripts/test-isolated.ts packages/server/src` 80/80 · `packages/ui` 1480 tests.

PR #755, opened 2026-08-21
fix(schedules): standard cron semantics, strict parsing, UTC alignment, timer persistence

Pizzaface:
Fixes the schedule-semantics findings from the Mode/Schedules/Triggers audit (Theme 4).

- Standard 5-field cron DOM/DOW OR semantics (was AND — schedules fired less often than documented).
- Strict cron token parsing (no `parseInt` prefix acceptance) and strict ISO time parsing with deterministic timezone handling.
- `HH:MM` frontend/backend alignment and explicit UTC labeling in schedule summaries.
- Relative timers persist absolute `fireAt` across runner restarts; overdue timers fire once on restart; missed cron occurrences catch up.
- Server-side validation of schedule params before persisting — invalid schedules are rejected instead of silently never arming.

Tests: UI suite 1490 passed / 1 skipped; root typecheck clean.

### DIFF
--- a/packages/cli/src/runner/services/time-utils.test.ts
+++ b/packages/cli/src/runner/services/time-utils.test.ts
@@ -279,9 +279,18 @@ describe("parseCron", () => {
         expect(parseCron("0 9 1x * *")).toBeNull();
     });
 
-    test("uses POSIX OR semantics for restricted DOM and DOW", () => {
-        const cron = parseCron("0 9 1 * 1")!;
-        expect(new Date(nextCronTime(cron, new Date("2026-06-02T10:00:00Z").getTime())!).toISOString()).toBe("2026-06-08T09:00:00.000Z");
+    test("uses POSIX day matching semantics", () => {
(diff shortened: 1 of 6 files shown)


Pizzaface:
!pizza - fix the checks

Pizzaface:
✅ Done!

**Changes:**
- Fixed POSIX cron day matching so wildcard DOM/DOW fields use AND semantics and two restricted fields use OR semantics.
- Added regression coverage for DOM-only, DOW-only, wildcard-step, and impossible schedules.
- Rebased onto current `main` and resolved the timer-state conflict in favor of main’s newer durability fixes.

**Verification:**
- Current-head Unit Tests: ✅ passing
- CodeQL: ✅ passing
- PR: ✅ mergeable
- Schedule service tests: ✅ 90 passing locally
- Schedule UI tests: ✅ 26 passing locally
- Typecheck: ✅ clean

Pushed rebased head `1a710eea` to `fix/schedule-semantics`. Downstream platform builds are now running.

PR #768, opened 2026-08-21
fix: align notification deep-link routing (C-003)

Pizzaface:
Night Shift dish C-003


Pizzaface:
Superseded by #836, which rebases this change onto main and resolves the App.tsx conflict with the restore-intent path.

PR #781, opened 2026-08-21
fix: reset per-session relay state on reconnect (A2-009)

Pizzaface:
Night Shift dish A2-009


Pizzaface:
Superseded by #844, which rebased this change onto latest main and includes the necessary test/cleanup fixes.

PR #803, opened 2026-08-21
fix: gate external PresentedCard images (C-006)

Pizzaface:
Night Shift dish C-006

## Problem
Model-provided schema.org image URLs in PresentedCard were rendered directly as ``, causing the viewer's browser to contact arbitrary attacker-controlled hosts on render — leaking IP, User-Agent, and timing.

## Fix
- External http(s) image URLs are now gated behind an explicit user click (click-to-load)
- Default render shows the kind icon (existing fallback) + a `title="Load image"` button
- Only after the user clicks does `` get set — no auto-fetch on render
- Non-external images (data:, relative) would render normally (currently httpImage only admits http/https, so all card.image values are gated)

## Changes
- `presented-card.ts`: Export `isExternalImage(url)` — returns true for http/https URLs
- `PresentedCard.tsx`: Add `imageApproved` state; external images default to kind icon + click button
- `presented-card.test.ts`: Unit tests for `isExternalImage`
- `PresentedCard.test.tsx`: Render tests — no `` on initial render, `` appears after click

## Verification
```
387 pass, 0 fail (bun test src/components/session-viewer)
tsc --noEmit: clean
```

### DIFF
--- a/packages/ui/src/components/session-viewer/PresentedCard.test.tsx
+++ b/packages/ui/src/components/session-viewer/PresentedCard.test.tsx
@@ -3,7 +3,7 @@
  */
 import { afterEach, describe, test, expect } from "bun:test";
 import { Window } from "happy-dom";
-import { render, cleanup } from "@testing-library/react";
+import { render, cleanup, fireEvent } from "@testing-library/react";
 import React from "react";
 import type { PresentedCard as PresentedCardData } from "./presented-card";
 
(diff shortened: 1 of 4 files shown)


Pizzaface:
superseded by #828 (merged) — external PresentedCard image gating with browser-normalized URL classification and URL-bound approval.

PR #815, opened 2026-08-21
fix: scope tunnel service messages across viewer switches (A2-015)

Pizzaface:
Night Shift dish A2-015


Pizzaface:
Superseded by #826, which rebases these changes onto latest main and resolves the App.test.tsx conflict with #817.

PR #821, opened 2026-08-21
fix: protect replacement sessions from stale cross-node disconnects (A2-017)

Pizzaface:
Night Shift dish A2-017

## Problem

On node B, registering a REPLACEMENT session cannot clear node A's LOCAL socket (socket maps are node-local). The stale node-A relay socket stays "valid"; when it LATER disconnects, its handler calls endSharedSession and ENDS/DELETES node B's replacement session. Stale events from the node-A socket are also still accepted (routed by socket token).

## Fix

**Shared connection-owner token** — the existing `token` field in Redis session state (already regenerated on every `registerTuiSession`) is used as the connection-generation owner token.

- **Disconnect handler** (`session-lifecycle.ts`): After the single-node local-socket guard, a new cross-node guard fetches `getSessionOwnerToken(sessionId)` from Redis. If the shared token differs from `socket.data.token`, this socket is stale/superseded — `endSharedSession` is **not** called.
- **Event pipeline** (`event-pipeline.ts`): Inside `enqueueSessionEvent`, the same token comparison runs before any state update or viewer broadcast. Stale events are silently dropped.

## Files Changed

- `packages/server/src/ws/sio-registry/sessions.ts` — `getSessionOwnerToken` helper (wraps single-field hGet)
- `packages/server/src/ws/sio-registry/index.ts` — export new helper
- `packages/server/src/ws/namespaces/relay/session-lifecycle.ts` — cross-node guard in disconnect handler
- `packages/server/src/ws/namespaces/relay/event-pipeline.ts` — cross-node guard in event enqueue
- `packages/server/src/ws/namespaces/relay/session-lifecycle.cross-node.test.ts` — two-node reconnect regression tests (4 cases)
- `packages/server/src/ws/namespaces/relay/event-pipeline.cross-node.test.ts` — event pipeline stale rejection tests (3 cases)
- `packages/server/src/ws/namespaces/relay/session-lifecycle.session-end.test.ts` — added missing mock for new export

## Verification

```
bun run typecheck && bun scripts/test-isolated.ts packages/server/src/ws
```

✅ Typecheck clean, 38/38 test files pass (0 failures)

### DIFF
--- a/packages/server/src/ws/namespaces/relay/child-lifecycle.ack.test.ts
+++ b/packages/server/src/ws/namespaces/relay/child-lifecycle.ack.test.ts
@@ -27,6 +27,8 @@ mock.module("../../sio-registry.js", () => ({
 }));
 
 mock.module("../../sio-state/index.js", () => ({
+    acquireSessionOwnershipLock: async () => {},
+    releaseSessionOwnershipLock: async () => {},
     removeChildSession: async (parentId: string, childId: string) => {
         childSets.get(parentId)?.delete(childId);
     },
--- a/packages/server/src/ws/namespaces/relay/event-pipeline.cross-node.test.ts
+++ b/packages/server/src/ws/namespaces/relay/event-pipeline.cross-node.test.ts
@@ -95,7 +95,11 @@ mock.module("@pizzapi/protocol", () => ({
 
 afterAll(() => mock.restore());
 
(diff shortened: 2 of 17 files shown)


Pizzaface:
superseded by #830 (merged) — atomic cross-node ownership validation during event processing and teardown.

PR #859, opened 2026-08-31
fix(cli): inject git branch into system prompt per session, not at boot

Pizzaface:
## Problem

Sessions always showed a stale git branch in their system prompt. The branch line was baked at worker boot (worker.ts / index.ts call buildSystemPrompt() inside the appendSystemPrompt IIFE), and the SDK caches that string for the life of the session. Any checkout after boot — dispatch machinery, the git panel, the agent itself — left the prompt claiming the old branch forever, and compaction faithfully preserved the lie.

## Fix

- Removed the boot-baked `Git branch:` line from `system-prompt.hbs` (and from the `SystemPromptContext` type generated by `scripts/compile-prompt.ts`). `gitWorktree` / `cwd` lines stay — they're stable for a session.
- New `git-branch` extension (`extensions/git-branch.ts`, registered in `factories.ts` for both worker and interactive CLI): on `before_agent_start`, looks up the branch with `git rev-parse --abbrev-ref HEAD` **once per session** — on the first turn, after any dispatch-time checkout — and injects the cached value above `Working directory: <cwd>` on every turn. Non-repo cwd → no line. Recomputes on `session_start` (in-process session switches).
- `runGit()` helper exported from `config/system-prompt.ts` for reuse.

Mid-session branch switches by the agent remain visible in conversation history, so the prompt keeps describing where the session started.

## Verification

- `tsc --noEmit` clean.
- New `git-branch.test.ts`: injection placement (above the cwd anchor, appended fallback), single-lookup memoization (stubbed `runGit`), session-switch reset, non-repo no-op. 39 targeted tests pass.
- Full-suite failure set compared against stashed baseline: **zero new failures**; 3 pre-existing failures fixed (two stale `gitBranch` assertions, one stale trigger-copy assertion that predates this branch).
- Smoke-tested end-to-end: baked prompt has no branch line; rendered prompt shows `Git branch: <live>` above `Working directory: <cwd>`.

### DIFF
--- a/packages/cli/scripts/compile-prompt.ts
+++ b/packages/cli/scripts/compile-prompt.ts
@@ -65,8 +65,6 @@ const spec = Handlebars.template(${precompiledSpec});
 export interface SystemPromptContext {
     /** Current date and time, e.g. "March 29, 2026, 3:45 PM" */
     dateTime: string;
-    /** Current git branch name, e.g. "feat/my-feature" */
-    gitBranch?: string;
     /** Git worktree root path (if running inside a worktree) */
     gitWorktree?: string;
     /** Current working directory */
--- a/packages/cli/src/config/system-prompt.test.ts
+++ b/packages/cli/src/config/system-prompt.test.ts
@@ -19,11 +19,6 @@ describe("buildSystemPrompt", () => {
         expect(result).toContain(year);
     });
 
-    test("includes gitBranch when provided", () => {
-        const result = buildSystemPrompt({ dateTime: "test", gitBranch: "feat/my-feature" });
-        expect(result).toContain("Git branch: feat/my-feature");
-    });
-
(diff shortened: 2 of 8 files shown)


Pizzaface:
!pizza - pls fix the tests

Pizzaface:
!pizza - fix the failing check.

Pizzaface:
✅ Done!

**Changes:**
- Corrected the stale trigger-prompt assertion in `packages/cli/src/config/system-prompt.test.ts` so it matches this branch’s actual template.

**Verification:**
- Targeted test: ✅ 21 passed
- Typecheck: ✅ clean
- Diff check: ✅ clean
- GitHub Unit Tests check: ✅ passed

Pushed commit `dbbe5bd5` to `bug/fix-stale-branch`.

PR #865, opened 2026-09-02
test(tools): adversarial sandbox tests

Pizzaface:
Extends `packages/tools/src/sandbox.test.ts` (+326 lines, 28 new passing tests, 3 `test.todo`) with adversarial cases against the real public API of the security boundary: deep/relative `../` traversal from inside allowed roots, prefix-confusion paths (`/tmp-evil` vs `/tmp`), URL-encoded segments (pinned as literal — the kernel never decodes `%2e%2e`), null bytes, empty/CWD-relative paths, symlinks escaping allowWrite/denyRead in both read and write directions (file and dir symlinks, plus a no-false-positive case for links resolving back inside), env-var leakage assertions (`getSandboxEnv` may only ever expose proxy-shaped `127.0.0.1` variables and must never carry `SSH_AUTH_SOCK`/AWS/secret values), network-config pass-through with edge-case domains (case, trailing dot, ports, wildcards — behavioral matching lives in srt and is not observable through this module's API), and malformed/empty config inputs. All tests use `mkdtempSync` under `tmpdir()` with `rmSync` cleanup, never touch `$HOME`, and are Linux-CI-safe (platform-conditional assertions where macOS sandbox init behaves differently). `bun test packages/tools` and `bun run typecheck` both exit 0.

**Bugs found (marked `test.todo`, `sandbox.ts` untouched per task scope):**

1. **`file://` URL prefix bypasses `denyRead`** — `validatePath("file:///etc/secrets/key", "read")` returns `{ allowed: true }` because `pathResolve` treats `file://...` as a CWD-relative path, so deny rules never match (reads fail open; writes fail closed only incidentally).
2. **`initSandbox` throws on malformed configs instead of degrading gracefully** — `{ mode: "basic" }` with `srtConfig` absent (undefined, not null) throws `TypeError` (`srt.network`), and undefined `denyRead`/`allowWrite`/`denyWrite` arrays throw during `_buildSrtConfig` (spread of undefined); both escape the documented "never crashes the worker" contract because guards only check `=== null` and `_buildSrtConfig` runs outside the try block.
3. **Dangling symlink under `allowWrite` passes write validation** — `existsSync` is false so the parent walk never resolves its target; an attacker who creates the symlink target between validation and write escapes the allowed root (TOCTOU window whenever OS-level enforcement is inactive).

### DIFF
--- a/packages/tools/src/sandbox.test.ts
+++ b/packages/tools/src/sandbox.test.ts
@@ -619,6 +619,18 @@ describe("sandbox", () => {
         });
 
         test("backslashes are literal filename characters on POSIX (no traversal)", async () => {
+            // Real-dir geometry that works on every POSIX platform — on macOS
+            // /etc is a symlink to /private/etc, which makes the /etc variant
+            // pass there by accident (rule resolves away from the literal
+            // path string). This sibling shape catches the bug everywhere.
+            const tmpDir = track(mkdtempSync(join(tmpdir(), "sandbox-bs-")));
+            await initSandbox(makeConfig({ denyRead: [join(tmpDir, "secrets")] }));
+            // <tmpDir>/secrets\x is a sibling named "secrets\x", NOT under
+            // <tmpDir>/secrets on POSIX — misread as secrets/x if separators
+            // were canonicalized outside Windows.
+            expect(validatePath(join(tmpDir, "secrets\\x"), "read").allowed).toBe(true);
+            expect(validatePath(join(tmpDir, "secrets", "x"), "read").allowed).toBe(false);
+            // The classic /etc variant pins it on Linux, where /etc is a real dir.
             await initSandbox(makeConfig({ denyRead: ["/etc"] }));
             // On POSIX backslash is not a separator, so this is one weird
             // filename, not /etc/secrets/x. Allowed — pinning correct POSIX
--- a/packages/tools/src/sandbox.test.ts
+++ b/packages/tools/src/sandbox.test.ts
@@ -685,15 +685,14 @@ describe("sandbox", () => {
             expect(result.reason).toContain("denied");
         });
 
-        test("dangling symlink under allowWrite passes validation (documents current gap)", async () => {
-            // existsSync() is false for a dangling symlink, so _normalizePath's
-            // parent walk treats it as an ordinary (unrealpath'd) child of the
-            // allowed root. The write itself would fail at the OS level today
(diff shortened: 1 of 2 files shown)


Pizzaface:
!pizza - fix the checks.

Pizzaface:
!pizza - fix the checks.

Pizzaface:
!pizza - fix the checks.

Pizzaface:
!pizza - fix this PR pls

Pizzaface:
✅ Done — pushed as [`5f9f251`](https://github.com/Pizzaface/PizzaPi/commit/5f9f2518c459d1160157035b68de9fc9e23f791c).

**Root cause:** `_toComparable()` in `packages/tools/src/sandbox.ts` replaced `\` with `/` on **every** platform. On POSIX the kernel treats `\` as a literal filename character, so a path like `/etc\secrets\x` (a file literally named `etc\secrets\x` sitting in `/`) was misread as `/etc/secrets/x` and wrongly tripped the `denyRead: ["/etc"]` rule. CI caught it on Linux because `/etc` is a real directory there; on macOS `/etc` is a symlink to `/private/etc`, which resolved the rule away from the literal path string and masked the bug locally — which is why the test went green on macOS and red on CI.

**Changes:**
- `packages/tools/src/sandbox.ts` — `_toComparable()` now canonicalizes separators only on Windows (`platform() === "win32"`), where `pathResolve` actually emits backslashes. Matches kernel semantics on both platforms: no deny-rule over-match on POSIX, no allow-check breakage on Windows.
- `packages/tools/src/sandbox.test.ts` — strengthened the existing pinning test with a symlink-free tmpdir geometry (`denyRead: <tmp>/secrets` vs sibling `<tmp>/secrets\x`) so the regression now fails on **every** POSIX platform, not just Linux. The classic `/etc` variant is kept.

**Verification:**
- Tools package tests: ✅ 194 pass, 0 fail (1 platform skip)
- Typecheck: ✅ clean across all packages
- Red/green: reverting the one-line fix makes the strengthened test fail locally; the fix makes it pass — confirmed the assertion actually catches the CI failure mode.
- CI: ✅ **Unit Tests passing** (previously the only red check). Remaining matrix jobs are builds that were already green on the prior commit.
- Note: `skills.test.ts` fails locally due to the known HOME-dependent environment issue (reads real `~/.pizzapi` files); unaffected by this diff and green on CI's restricted `$HOME`.

PR #866, opened 2026-09-02
chore: add oxlint to CI

Pizzaface:
Adds [oxlint](https://oxc.rs/) as the repo linter: dev dependency + root `lint` script (`oxlint packages/*/src`), a minimal `.oxlintrc.json` enabling the `correctness` category as errors, a `Lint` step in CI next to Typecheck, and `bun run lint` documented in AGENTS.md.

**Green baseline.** The unconfigured default run produced 330 warnings; rather than rewrite the codebase, the config disables 12 rules (all findings listed by count): `no-unused-vars` (217 hits / 126 files), `no-useless-escape` (38), `no-control-regex` (23, mostly intentional ANSI-escape and JSON-control-char regexes), `no-useless-spread` (8, removing it can change live-iteration semantics), `no-unsafe-optional-chaining` (8, needs real guards), `prefer-string-starts-ends-with` (2), `no-self-assign` (2, test fixtures), `no-invalid-fetch-options` (2, the flagged `new Request(req, { body })` in handler.ts is an intentional workaround for proxied GET bodies — see its comment), `no-useless-length-check` (1), `no-unused-private-class-members` (1), `no-unused-expressions` (1), `no-irregular-whitespace` (1).

**26 mechanical fixes applied** (provably identical semantics): 18 empty object-spread fallbacks `{...(x ?? {})}` → `{...x}` (spread of null/undefined is a no-op) and 8 `new Array(n)` → `Array(n)`.

Gates: `bun run lint` exit 0, `bun run typecheck` exit 0, `bun run test` passes except the pre-existing, documented HOME-dependent `skills.test.ts` failure on dev machines with global `~/.pizzapi/rules` (verified passing with an empty HOME, as in CI; untouched by this diff).

## ⚠️ Behavior change (not pure lint)

`packages/cli/src/extensions/mcp/registry.ts`: streamable-transport MCP servers now receive `expandVarsDeep(s.headers)` instead of the raw `s.headers`. Previously `@VAR@` tokens in streamable headers were silently left unexpanded (http transport was already correct). Surfaced by `no-unused-vars` flagging the discarded `headers` local. Regression test: `mcp.test.ts` → "streamable transport expands @VAR@ tokens in headers".

All 12 initially-disabled rules are now re-enabled — see the review comment below for the per-rule table.


Pizzaface:
All 12 disabled rules are now re-enabled — every finding fixed or explicitly suppressed with a reason. One commit per rule: 781f6c01, e67216a7, 8d675bfa, 0b49437e, 12cbe5a7, dfa1049e.<br /><br /><table>
<tr><th>Rule</th><th>Hits fixed</th><th>Status</th></tr>
<tr><td>no-unused-vars</td><td>217</td><td>enabled — ignore patterns ^_ for args/vars/caught; 132 unused imports removed (ts-morph codemod), dead locals/vars removed, dead private code deleted, intentionally-unused params prefixed _; also wired a latent bug (streamable MCP headers were computed but never passed)</td></tr>
<tr><td>no-useless-escape</td><td>38</td><td>enabled — backslash removals in regex literals / char classes / template literal; regex rewrites verified semantics-identical</td></tr>
<tr><td>unicorn/no-useless-spread</td><td>8</td><td>enabled — 7 provably identical removals; 1 inline disable in git-service notifyRepoChanged (deliberate snapshot: awaits inside the loop allow concurrent map mutation)</td></tr>
<tr><td>no-self-assign</td><td>2</td><td>enabled — no-op globalThis.SyntaxError self-assignments removed from test preambles</td></tr>
<tr><td>no-useless-length-check</td><td>1</td><td>enabled — redundant ids.length guard dropped (some() on empty is false)</td></tr>
<tr><td>no-unused-private-class-members</td><td>1</td><td>enabled — write-only #socket field removed from TimeService</td></tr>
<tr><td>no-unused-expressions</td><td>1</td><td>enabled — kill() ternary statement converted to if/else</td></tr>
<tr><td>no-irregular-whitespace</td><td>1</td><td>enabled — zero-width space in a cron doc comment replaced with a regular space</td></tr>
<tr><td>unicorn/prefer-string-starts-ends-with</td><td>2</td><td>enabled — startsWith over caret regexes</td></tr>
<tr><td>no-unsafe-optional-chaining</td><td>8</td><td>enabled — all test assertions guarded so a missing parent fails the assertion instead of throwing TypeError</td></tr>
<tr><td>no-control-regex</td><td>23</td><td>enabled — code unchanged; each site (ANSI stripping, git ref validation, control-char assertions, ByteString validation) carries an oxlint-disable-next-line with reason</td></tr>
<tr><td>unicorn/no-invalid-fetch-options</td><td>2</td><td>enabled — code unchanged; both handler.ts sites are the documented proxied-GET-body workaround, inline disable with reason</td></tr>
</table><br /><br />Exit codes: bun run lint = 0, bun run typecheck = 0, bun run test = 184 passed / 1 failed — the only failure is the pre-existing HOME-dependent packages/cli/src/skills.test.ts (5 cases, unchanged by this PR). Two unrelated flakes were observed intermittently and reproduced on the base commit as well (runner-recent-folders.test.ts prune test, git-service.state.test.ts external-checkout test — git index.lock contention under parallel workers); both pass standalone and are untouched by this diff.

PR #872, opened 2026-09-03
fix(server): route filter eq is case-insensitive for strings

Pizzaface:
A github:pr_comment route filtering author eq "pizzaface" never matched
GitHub's "Pizzaface" login, so every event was published and silently
dropped. GitHub logins, repos and branches are case-insensitive; make
string eq match the same way (contains already does). Non-string values
keep loose equality.

### DIFF
--- a/packages/cli/src/extensions/triggers/extension.ts
+++ b/packages/cli/src/extensions/triggers/extension.ts
@@ -874,13 +874,14 @@ export const triggersExtension: ExtensionFactory = (pi) => {
                 },
                 filters: {
                     type: "array",
-                    description: "Optional delivery filters based on the trigger's output schema. Each filter is { field, value, op? }. 'field' is a payload property name, 'value' is the expected value (or array for OR), 'op' is 'eq' (default) or 'contains'.",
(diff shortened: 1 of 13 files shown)


Pizzaface:
!pizza - fix the checks

Pizzaface:
✅ Fixed — pushed `a5625bb6` + `0dbf32b0` to `fix/filter-eq-case-insensitive`.

**Root cause:** not from this branch's changes. PR #871 added named imports of `getAuthContext` / `runWithAuthContext` from `auth.js` to `events/transport.ts`, but several test files' `mock.module("../auth.js")` factories only exported `getKysely`. Bun fails at module-link time when a mocked module lacks a named import, so any test whose import graph reached `transport.ts` died with `Export named "getAuthContext" not found`.

**Broken files (all pre-existing on main from #871):**
- `routes/events.test.ts` — 2 failures
- `routes/runner-owner.test.ts` — 1 failure
- `routes/webhooks.test.ts` — 1 failure (hidden: `test-isolated` headline counts didn't surface it)

**Fix:** supplied the two missing exports in all three test mocks (dummy context + direct callback invocation — none of these tests exercise the acked emit path). No production code changed.

**Verification:**
- All three files pass locally (85 / 31 / 54 tests)
- Full `test-isolated.ts packages/server/src` run: only remaining local failures are `ECONNREFUSED 127.0.0.1:6379` in 3 Redis-dependent files that pass in CI (Redis service) — environmental
- `bun run typecheck`: clean

**Heads-up:** main is currently red from this same breakage (CI failed on the #871 merge commit) — merging this PR should turn it green again.

